PR #115, opened 2013-11-24
extension in setup script added

rhoef:
setup script does not build the extension. 
- removed setuptools, using pythons distutils instead
- no need for cmake to build extension drtile.so, see changes in installation.rst for details


k-dominik:
@stuarteberg : Do you, by any chance, know if this is still useful?

chaubold:
I thought we completely got rid of `drtile` anyways?

stuarteberg:
Right, we there is no extension module in lazyflow any more.

PR #152, opened 2014-10-17
Changed setValue for the special case of value is None

jakirkham:
If value is None, then this should have the effect of making the slot unready. In addition, notifications should be sent if the state changed. These will be either ready or unready notifications based on the value given.

### DIFF
--- a/lazyflow/slot.py
+++ b/lazyflow/slot.py
@@ -1099,12 +1099,15 @@ class Slot(object):
                 for s in self._subSlots:
                     s.setValue(self._value)
     
-                notify = (self.meta._ready == False)
-    
-                # a slot with a value is always ready
-                self.meta._ready = True
-                if notify:
-                    self._sig_ready(self)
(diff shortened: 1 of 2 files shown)


jakirkham:
I should add that using setValue with None previously had undefined behavior. In other words, it raised an exception and left the slot in an undefined state. This fixes that problem.

jakirkham:
Rebased on current master.

jakirkham:
Added tests to verify that setting an `InputSlot` to `None` works.

Note that these tests would have failed previously due to indeterminate behavior. In other words, setting the value would work; however, the slot would still claim to be ready and would still try to retrieve a value and fail with the message below.

    AssertionError: This inputSlot has no value and no partner.  You can't ask for its data yet!`

In other words, this pull request not only provides a feature, but fixes a serious bug.

jakirkham:
@stuarteberg, can you please look at this again when you have a chance?

PR #158, opened 2014-12-03
[ready] Adding laziness to OpFilterLabels

burgerdev:
There's also a new parent operator for general region growing which ~~could be~~ is used by `OpLabelVolume`, too.


burgerdev:
partial implementation for ilastik/ilastik#1020

k-dominik:
@akreshuk you know the history here. As now would be a natural timepoint to decide what to do with this PR. What is your take on it? Can we scratch it right away, should look into it more before deciding or work to merge it soon?

burgerdev:
It does not seem prudent to merge code that old without demand and maintainer. If somebody wants to keep working on this I will try my best to hand it over, but tbh I'm not that deep into the code myself after all those years.

k-dominik:
thx @burgerdev for the assessment and taking action

PR #160, opened 2014-12-16
[READY] fixes ilastik/ilastik#1050 and other caching issues

### LINKED ISSUE #79: Deadlocks in ArrayCache memory management code
The ArrayCacheMemoryMgr class and OpArrayCache work cooperatively to implement an LRU cache replacement strategy.  This code needs review and thorough testing.  Deadlocks inevitably occur during long memory-intensive batch jobs.  An inspection of the code reveals multiple opportunities for lock order inversion to occur, involving ArrayCacheMemoryMgr._lock, OpArrayCache._lock, and OpArrayCache._cacheLock, which are presumably the source of the deadlocks.

### LINKED ISSUE #148: [OpArrayCache] stuff going wrong during report generation
@thorbenk (git blame result)

Generation of cache reports fails when block shapes are tuples and not ndarrays because they do not support arithmetic operations. Furthermore, the function in which the error occurs does seem to calculate the actual block shape wrong and does not return a value.

OpArrayCache.py @ line 102:
```python
def _blockShapeForIndex(self, index):
    if self._cache is None:
        return None
    cacheShape = numpy.array(self._cache.shape)
    blockStart = index * self._blockShape
    blockStop = numpy.minimum(blockStart + self._blockShape, cacheShape)
```

Should be something like:
```python
def _blockShapeForIndex(self, index):
    if self._cache is None:
        return None
    cacheShape = numpy.array(self._cache.shape)
    blockShape = numpy.array(self._blockShape)
    blockStart = index*blockShape  # ??? possibly incorrect
    blockStop = numpy.minimum(blockStart + blockShape, cacheShape)
    return blockStop - blockStart  # correct return value?
```

burgerdev:
Issue ilastik/ilastik#1050 fixed, but the report generation does still not do what is expected, see #148.

fixes #79 
fixes #148

### DIFF
--- a/lazyflow/operators/opArrayCache.py
+++ b/lazyflow/operators/opArrayCache.py
@@ -100,11 +100,7 @@ class OpArrayCache(Operator, ManagedCache):
     # ========== CACHE API ==========
 
     def usedMemory(self):
-        try:
-            return self._cache.nbytes
-        except:
-            # cache is empty or deleted, or somehting along these lines
-            return 0
+        return self._usedMemory(self._cache)
 
     def fractionOfUsedMemoryDirty(self):
         if self.Output.meta.shape is None:
@@ -138,7 +134,37 @@ class OpArrayCache(Operator, ManagedCache):
         else:
             return 0
 
+    def generateReport(self, memInfoNode):
+        super(OpArrayCache, self).generateReport(memInfoNode)
+        if self._cache is not None:
+            if hasattr(self._cache, "dtype"):
+                memInfoNode.dtype = self._cache.dtype
+            else:
+                # cache is no array, so we cannot determine the dtype
+                pass
+
     # ========== END CACHE API ==========
+
+    @staticmethod
+    def _usedMemory(item):
+        s = 0
+        if isinstance(item, numpy.ndarray):
+            if item.dtype == numpy.object:
+                for x in item.ravel():
+                    s += OpArrayCache._usedMemory(x)
+            else:
+                s = item.nbytes
+        elif isinstance(item, dict):
+            for key in item.keys():
+                try:
+                    obj = item[key]
+                except KeyError:
+                    # cleaned up
+                    pass
+                else:
+                    s += OpArrayCache._usedMemory(obj)
+        return s
+                    
 
     def _blockShapeForIndex(self, index):
         if self._cache is None:
--- a/lazyflow/operators/opBlockedArrayCache.py
+++ b/lazyflow/operators/opBlockedArrayCache.py
@@ -204,7 +204,7 @@ class OpBlockedArrayCache(Operator, ObservableCache):
         report.type = type(self)
         report.id = id(self)
 
-        for block_index, block in self._cache_list.iteritems():
+        for block_index in self._cache_list.keys():
             start = self._blockShape*self._get_block_multi_index(block_index)
             stop = map(lambda z: z[0]+z[1], zip(start, self._blockShape))
             stop = numpy.minimum(stop, self.Output.meta.shape)
@@ -212,7 +212,13 @@ class OpBlockedArrayCache(Operator, ObservableCache):
             n = MemInfoNode()
             n.roi = (start, stop)
             report.children.append(n)
-            block.generateReport(n)
+            try:
+                block = self._cache_list[block_index]
+            except KeyError:
+                # probably cleaned up 
+                pass
+            else:
+                block.generateReport(n)
             
     def usedMemory(self):
         tot = 0.0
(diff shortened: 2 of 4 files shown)


burgerdev:
also fixes ilastik/ilastik#807 (the part wjere memory calculation fails).

stuarteberg:
Thanks for working on this.  This is cleaner than the existing implementation.  I have a couple of remarks:

1. In the old implementation, we attempted to prioritize cache removal with an LRU heuristic.  As you discussed in #173, this calculation is not working properly now, and maybe never quite worked correctly.  In this new branch, the LRU scheme has been removed entirely.  Can we take a crack at adding it back? (And try to get it right this time? :-)

2. Can we modify the cache interface to allow partially freeing a cache?  In the current implementation, the entire contents of a cache are freed at once.  That's OK for caches like `OpBlockedArrayCache`, since each block is stored as an internal `OpArrayCache` -- individual blocks can be freed independently because they are caches in their own right.  But some caches like `OpCompressedCache` and `OpUnblockedArrayCache` manage their data directly.  It would be nice if we could reclaim *some* of their blocks without reclaiming all of it.

burgerdev:
Thanks for looking into it :)

  1. I stumbled over the term 'LRU scheme' while rewriting the manager, and I have to admit that I did not quite get what was meant there. [Wikipedia](http://en.wikipedia.org/wiki/Cache_algorithms#Examples) indicates that LRU is simply 'discard the least recently used cache first', which I hope I am doing [by sorting the caches](https://github.com/burgerdev/lazyflow/blob/ilastik-issue-1050/lazyflow/operators/cacheMemoryManager.py#L206). Did I miss something?

  2. This sounds like a good idea. The most simple addition I can think of is `OpDirtyCache.freeDirtyMemory()`. A fine grained control over ROIs to free would be more interesting, but I'm not sure how to implement this -- how would the manager decide which ROIs are safe to delete? We could extend the `lastAccessTime()` method to accept a ROI, or to return a map of ROIs and timestamps.

stuarteberg:
>LRU is simply 'discard the least recently used cache first', which I hope I am doing by sorting the caches. Did I miss something?

No, I did.  I somehow missed that line.  Thanks for clarifying.

>The most simple addition I can think of is `OpDirtyCache.freeDirtyMemory()`.

That's a good idea: We should prioritize clearing dirty blocks first.

>A fine grained control over ROIs to free would be more interesting, but I'm not sure how to implement this -- how would the manager decide which ROIs are safe to delete? 

I think that decision should be left to the caches themselves.

>We could extend the `lastAccessTime()` method to accept a ROI, or to return a map of ROIs and timestamps.

The latter makes sense to me (but using arbitrary `block_id`s instead of ROIs). Perhaps something like this would work:

```python
# we need a cache cleanup
caches = list(self._managed_caches)

# Construct a heap of all cache blocks, sorted by access time.
import heapq
block_heap = []

for cache in caches:
    for timestamp, block_id in cache.getBlockAccessTimes():
        heapq.heappush( block_heap, (timestamp, cache, block_id) )

while current_usage_percentage > self._target_usage and block_heap:
    timestamp, cache, block_id = heapq.heappop(block_heap)
    self.logger.debug("Cleaning up cache: '{}', block: {}".format(cache.name, block_id))
    cache.freeBlock(block_id)
    current_usage_percentage = memoryUsagePercentage()

# don't keep a reference until next cleanup!
del cache
```

Note that the code above requires new functions for `OpCache`:
- `getBlockAccessTimes()` instead of `lastAccessTime()`
- `freeBlock()` instead of `freeMemory()`

stuarteberg:
On a different subject: I'm wondering if `OpCache` might be easier to use in some cases if it were a "mixin" instead of a regular base class.  It's a subtle distinction, but still.  What do you think?

burgerdev:
Nice, let the caches to the difficult stuff! I will go test this right away. 

Regarding the mixin: do you mean that we should separate `Operator` from `Cache`?

stuarteberg:
>Regarding the mixin: do you mean that we should separate `Operator` from `Cache`?

Genau.  There is one minor trick to this, because you'll have to essentially re-implement the `after_init()` mechanism in the MixIn class.  (I don't think you can rely on developers inheriting from `Operator` and `OpCache` in a consistent order, so you can make assumptions about the `__mro__`.) But it shouldn't be hard.

burgerdev:
@stuarteberg: I think I am done implementing stuff here, we should test this in ilastik now.

stuarteberg:
I'm ready to merge this if you are.  Should I merge your `memUsage2` ilastik branch simultaneously?

burgerdev:
I am still testing the whole PR in ilastik: there were already some bugs, so I would like to keep looking around a bit to see if I can find more. When I'm done I will change the title and write another comment here. 

We have to merge the two branches (this one and `memUsageDialog2`) simultaneously because `ArrayCacheMemoryMgr` was deleted from this branch.

burgerdev:
I guess this is ready to merge.

Issue #162, opened 2015-02-03
Be able to handle frames of variable size (with known maximum)


jakirkham:
See full text here ( https://github.com/ilastik/ilastik/issues/1076 ).

PR #163, opened 2015-02-03
Masked arrays i.e. Handle frames of variable size (with known maximum)

jakirkham:
Provides support for masked arrays in lazyflow. This is done by adding `has_mask` to the slot metadata. This value defaults to `False`, which provides the default behavior that ilastik and lazyflow are accustomed. When `has_mask` is set to True, lazyflow will allocate a masked array instead of a regular NumPy ndarray. They behave the same, but the masked array allows for hiding or ignore values that are considered erroneous, which allows for processing a smaller portion of image stack, whose shape cannot be known until execution is run. This is related to several issues ( https://github.com/ilastik/ilastik/issues/1076 ) (  https://github.com/ilastik/lazyflow/issues/162 ) ( https://github.com/ilastik/volumina/issues/151 ).


jakirkham:
A complimentary pull request for volumina can be seen as well ( https://github.com/ilastik/volumina/pull/152 ).

lazyflow/metaDict.py
```diff
@@ -40,6 +40,9 @@ def __init__(self, other=None, *args, **kwargs):
             # are ready
             self._ready = False
 
+        if not 'has_mask' in self:
+            self.has_mask = False
+
```
stuarteberg: Very minor change needed here: In the `MetaDict` class, all attributes that are not present in the dict are already returned as `None` if someone attempts to access them.  Since `has_mask` only applies to array data, and it defaults to `False` anyway, I think it would be easier to simply avoid initializing it with a value here.  Just delete these two lines.  Then `has_mask` will be implemented in the same way as other array-like slot metadata (e.g. `shape`, `dtype`, `axistags`, etc.)

stuarteberg:
These changes look good to me.  Thanks for implementing test cases.  I requested one minor change in on the diff page.  After it is fixed, we can merge this.

lazyflow/metaDict.py
jakirkham: I did notice that keys/attributes that weren't present would return `None`. However, as the key/attribute name is `has_mask`, I thought it best to make it unambiguously boolean. This way the "end user" would not have any surprises by making the naive assumption that it is strictly boolean. As this was done with `_ready`, I did not think this was an issue. If you still think it should be changed, I would be happy to do so.

jakirkham:
Thanks for reviewing this. I tried to give you some background on my design choice that you have commented on. Please let me know your thoughts. In the meantime, don't merge this yet I have some more changes I wanted to add.

lazyflow/metaDict.py
stuarteberg: I'd still prefer to change it.  It isn't obvious here, but `_ready` is a very special case -- it's not considered part of this class's "public" attributes.  But the new `has_mask` attribute isn't special -- it is very analogous to other fields like `prefer_2d`, among others.

jakirkham: Alright, I will change it.

jakirkham:
I have made some changes to the content in this PR particular with regards to `tests/testMaskedArray.py`. First, I noted that `OpMaskArrayIdentity` is basically the same as `OpArrayPiper`. So, I replaced the former with the latter. Second, I noticed that `OpMaskArrayBorder` could be refactored into a usable operator of its own, which I have termed `OpMaskArray`. As a consequence of this, the tests from `tests/testMaskedArray.py` got moved into separate files for each operator. Finally, I responded to your comments. Please let me know your thoughts.

stuarteberg:
Looks good to me.  Thinking about these changes a bit further, I'm wondering what we should do (if anything) about the fact that we're now "overloading" the word "mask" in some places.   For instance, there is already an operator named [`opMaskedSelect`][1] which zeros-out pixels outside of a mask.  Another example is [`opMaskedWatershed`][2], which performs a watershed operation on an image within a mask boundary.  In both of these cases, the "mask" is accepted as a totally separate input slot, not as a masked array.

Do you think we need to come up with some special terminology to distinguish between these cases?  Or perhaps re-write those operators to use masked arrays?

[1]: https://github.com/ilastik/lazyflow/blob/master/lazyflow/operators/opMaskedSelect.py
[2]: https://github.com/ilastik/lazyflow/blob/master/lazyflow/operators/opMaskedWatershed.py

stuarteberg:
In any case, I think my comments above do not need to be resolved for this PR to be merged.  Merging now.

jakirkham:
Fair questions, I can take a look at those operators and come up with some thoughts about how we might address this ambiguity.

PR #164, opened 2015-02-12
Masked arrays i.e. Handle frames of variable size (with known maximum) (pt. 2)

jakirkham:
A continuation of the work done in part 1 ( https://github.com/ilastik/lazyflow/pull/163 ). This work is here to address these outstanding issues ( https://github.com/ilastik/ilastik/issues/1076 ) ( https://github.com/ilastik/lazyflow/issues/162 ) ( https://github.com/ilastik/volumina/issues/151 ).

Makes a few minor changes in lazyflow to ensure masked arrays are copied correctly as there is a bug in NumPy ( https://github.com/numpy/numpy/issues/5558 ). In addition, various caches were modified to ensure that they can properly handle masked arrays. This includes opCompressedCache, OpCompressedUserLabelArray, OpArrayCache, OpBlockedArrayCache, and OpSlicedBlockedArrayCache. Other operators can provide support by following similar strategies. Tests, using masked arrays, were provided for all of these caches. In some cases, test did not exist or were not stringent enough and have since been added or tightened up.

### DIFF
--- a/lazyflow/operators/opCompressedUserLabelArray.py
+++ b/lazyflow/operators/opCompressedUserLabelArray.py
@@ -54,17 +54,17 @@ class OpCompressedUserLabelArray(OpCompressedCache):
     nonzeroBlocks = OutputSlot()
     #maxLabel = OutputSlot()
     
-    Projection2D = OutputSlot() # A somewhat magic output that returns a projection of all 
-                                # label data underneath a given roi, from all slices.
-                                # If, for example, a 256x1x256 tile is requested from this slot,
-                                # It will return a projection of ALL labels that fall within the 256 x ... x 256 tile.
-                                # (The projection axis is *inferred* from the shape of the requested data).
-                                # The projection data is float32 between 0.0 and 1.0, where:
-                                # - Exactly 0.0 means "no labels under this pixel"
-                                # - 1/256.0 means "labels in the first slice"
-                                # - ...
-                                # - 1.0 means "last slice"
-                                # The output is suitable for display in a colortable.
+    Projection2D = OutputSlot(allow_mask=True) # A somewhat magic output that returns a projection of all
+                                               # label data underneath a given roi, from all slices.
+                                               # If, for example, a 256x1x256 tile is requested from this slot,
+                                               # It will return a projection of ALL labels that fall within the 256 x ... x 256 tile.
+                                               # (The projection axis is *inferred* from the shape of the requested data).
+                                               # The projection data is float32 between 0.0 and 1.0, where:
+                                               # - Exactly 0.0 means "no labels under this pixel"
+                                               # - 1/256.0 means "labels in the first slice"
+                                               # - ...
+                                               # - 1.0 means "last slice"
+                                               # The output is suitable for display in a colortable.
     
     def __init__(self, *args, **kwargs):
         super(OpCompressedUserLabelArray, self).__init__( *args, **kwargs )
--- a/lazyflow/operators/opCompressedUserLabelArray.py
+++ b/lazyflow/operators/opCompressedUserLabelArray.py
@@ -144,21 +144,7 @@ class OpCompressedUserLabelArray(OpCompressedCache):
         for block_roi in stored_block_rois:
             # Get data
             block_shape = numpy.subtract( block_roi[1], block_roi[0] )
-            block = numpy.ndarray( shape=block_shape, dtype=self.Output.meta.dtype )
-
-            if self.Output.meta.has_mask:
-                block_fill_value = None
-                if issubclass(block.dtype.type, numpy.integer):
-                    block_fill_value = block.dtype.type(numpy.iinfo(block.dtype.type).max)
-                elif issubclass(block.dtype.type, numpy.floating):
-                    block_fill_value = block.dtype.type(numpy.nan)
-
-                block = numpy.ma.masked_array(
-                    block,
-                    mask=numpy.ma.getmaskarray(block),
-                    fill_value=block_fill_value,
-                    shrink=False
-                )
+            block = self.Output.stype.allocateDestination(SubRegion(self.Output, *roiFromShape(block_shape)))
 
             self.execute(self.Output, (), SubRegion( self.Output, *block_roi ), block)
 
@@ -391,21 +377,7 @@ class OpCompressedUserLabelArray(OpCompressedCache):
         """
 
         # Extract the data to modify
-        original_data = numpy.ndarray( shape=new_pixels.shape, dtype=self.Output.meta.dtype )
-        # Use masked array if that is what Output expects.
-        if self.Output.meta.has_mask:
-            original_data_fill_value = None
-            if issubclass(original_data.dtype.type, numpy.integer):
-                original_data_fill_value = original_data.dtype.type(numpy.iinfo(original_data.dtype.type).max)
-            elif issubclass(original_data.dtype.type, numpy.floating):
-                original_data_fill_value = original_data.dtype.type(numpy.nan)
-
-            original_data = numpy.ma.masked_array(
-                original_data,
-                mask=numpy.ma.getmaskarray(original_data),
-                fill_value=original_data_fill_value,
-                shrink=False
-            )
+        original_data = self.Output.stype.allocateDestination(SubRegion(self.Output, *roiFromShape(new_pixels.shape)))
 
         self.execute(self.Output, (), roi, original_data)
         
--- a/lazyflow/stype.py
+++ b/lazyflow/stype.py
@@ -110,7 +110,7 @@ class ArrayLike( SlotType ):
             storage_mask = numpy.zeros(storage.shape, dtype=bool)
             storage_fill_value = None
             if issubclass(storage.dtype.type, numpy.integer):
-                storage_fill_value = storage.dtype.type(numpy.iinfo(storage.dtype.type).max)
+                storage_fill_value = storage.dtype.type(0)
             elif issubclass(storage.dtype.type, numpy.floating):
                 storage_fill_value = storage.dtype.type(numpy.nan)
             elif issubclass(storage.dtype.type, numpy.complexfloating):
--- a/lazyflow/stype.py
+++ b/lazyflow/stype.py
@@ -127,6 +127,14 @@ class ArrayLike( SlotType ):
         return storage
 
     def writeIntoDestination( self, destination, value, roi ):
+        # If we do not support masked arrays, ensure that we are not being passed one.
+        assert self.slot.allow_mask or (not self.slot.meta.has_mask), \
+            "A masked array was provided as a destination. However," \
+            " the slot, \"%s\", of operator, " "\"%s\", does not support masked arrays." \
+            " If you believe this message to be incorrect, " \
+            "please pass the keyword argument `allow_mask=True` to the slot constructor." \
+            % (self.slot.operator.name, self.slot.name)
+
         if destination is not None:
             if not isinstance(destination, list):
                 assert(roi.dim == destination.ndim), "%r ndim=%r, shape=%r" % (roi.toSlice(), destination.ndim, destination.shape)
--- a/lazyflow/stype.py
+++ b/lazyflow/stype.py
@@ -20,6 +20,7 @@
 #		   http://ilastik.org/license/
 ###############################################################################
 import numpy, vigra
+import collections
 import warnings
 
 from roi import roiToSlice
@@ -200,7 +201,8 @@ class ArrayLike( SlotType ):
             dst.mask[...] = numpy.ma.getmaskarray(src_val)
             if isinstance(src_val, numpy.ma.masked_array):
                 dst.fill_value = src_val.fill_value
-        elif isinstance(dst, (list, tuple)) or isinstance(src, (list, tuple)):
+        elif isinstance(dst, collections.MutableSequence) or \
+                isinstance(src, collections.MutableSequence):
             dst[:] = src[:]
         else:
             dst[...] = src[...]
--- a/lazyflow/stype.py
+++ b/lazyflow/stype.py
@@ -109,7 +109,7 @@ class ArrayLike( SlotType ):
         if self.slot.meta.has_mask:
             storage_mask = numpy.zeros(storage.shape, dtype=bool)
             storage_fill_value = None
-            if issubclass(storage.dtype.type, numpy.integer):
+            if issubclass(storage.dtype.type, (numpy.bool8, numpy.integer)):
                 storage_fill_value = storage.dtype.type(0)
             elif issubclass(storage.dtype.type, numpy.floating):
                 storage_fill_value = storage.dtype.type(numpy.nan)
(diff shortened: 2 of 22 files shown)


jakirkham:
@stuarteberg Please let me know your thoughts.

lazyflow/operators/opCompressedUserLabelArray.py
```diff
@@ -247,12 +262,24 @@ def _executeProjection2D(self, roi, destination):
...
+                deep_data = block[block_relative_intersection_slicing]
+
+            # make binary and convert to float (must copy)
+            deep_data_float = deep_data.astype(numpy.float32)
+            deep_data_float[deep_data_float.nonzero()] = 1
 
-            # make binary and convert to float
-            deep_data_float = numpy.where( deep_data, numpy.float32(1.0), numpy.float32(0.0) )
-            
             # multiply by slice-index
```
stuarteberg: Interesting.  I thought replacing `numpy.where()` in this manner would be slower since (1) you can't copy/update in one-step and (2) `deep_data_float.nonzero()` creates a temporary array that wasn't needed previously.

So, I just did a quick experiment.  It turns out that your change is ~3x slower for non-sparse data (i.e. few nonzero pixels), presumably due to the huge size of the `nonzero()` coordinates in that case.  But in this function, the data will generally be quite sparse, in which case your change is ~2x faster.

jakirkham: What size data and type did you use?

If we decide it needs to be `where`, there is `bumpy.ma.where`, which returns a masked array that could be used in a simple branch.

stuarteberg: I used data of shape `(100,1000,1000)`, with `deep_data.dtype == uint8`.  But in this particular case (sparse data), your change is better.

>If we decide it needs to be where, there is numpy.ma.where, which returns a masked array that could be used in a simple branch.

Wait, so... are you saying that `numpy.where()` can't be used in lazyflow any more?  Perhaps we need to define a helper function that transparently handles both cases (masked/unmasked).  Or maybe even monkey-patch `numpy.where()`?

What other ordinary numpy functions will we need to avoid now that we want to support masked arrays?

jakirkham: > Wait, so... are you saying that numpy.where() can't be used in lazyflow any more? Perhaps we need to define a helper function that transparently handles both cases (masked/unmasked). Or maybe even monkey-patch numpy.where()?
> 
> What other ordinary numpy functions will we need to avoid now that we want to support masked arrays?

That is a very interesting question.

If we want to generalize the use of masked arrays further in the pipeline, we need to think about what they mean in the context of other operators. At this point, I haven't tried to address this question generally as it is not clear to me for all operators what a mask should mean. Certainly, I have some thoughts on this, but am interested in hearing others. I wouldn't want to force anybody into a particular definition until we at least had some discussion.

NumPy generally uses the mask to indicate erroneous regions or regions that should be ignored for calculation. For instance, division by zero results in masking and calculating means completely exclude masked values. Matplotlib and qimage2ndarray treat the mask as transparent (i.e. ignored). In Volumina, we are now consistently following the latter practice (it was already followed to some extent before just not consistently). It’s not clear to me what practice we want to follow in Lazyflow. We could follow the NumPy practice exactly, or we could only ignore values, or we could have some other use cases in mind.

As a result, I tried to steer clear of answering this question by focusing only on a subset of cache operators, which were intended, despite some sophisticated internals, to behave as identity operators. In this case, I would expect a masked array to propagated through intact. However, I did tweak `OpCompressedUserLabelArray` to propagate the mask and this is not exactly an identity operator. Another general principle I tried to follow is that a `numpy.ndarray` should behave as a masked array with a mask of `False`.

As far as the internals are concerned, we would want to check whether to use `numpy.where` or `numpy.ma.where`. Unfortunately, `numpy.ma.where` on normal `numpy.ndarray`s results in a masked array. I agree with you that we would want to avoid having people do this themselves. So, having some helper functions in the interim is a good idea. This is why I have made changes to `lazyflow.stype` and have tried to more generally use methods from there in the operators that I changed. In the case of `numpy.where`, I think a PR is in order here and I would be happy to do it. Though `numpy.where` is an exception, not the norm. Using `numpy.asarray` is bad. It is better to use `numpy.asanyarray`. Functions from `numpy` normally work fine with masked arrays. Most of the work here was just making sure the proper array was allocated when needed.

jakirkham: > I used data of shape (100,1000,1000), with deep_data.dtype == uint8. But in this particular case (sparse data), your change is better.

Cool. Just to verify, were you copying the data back into the original array ( [L291](https://github.com/ilastik/lazyflow/pull/164/files#diff-f2aeda246096a7282ca7ee818a962302L291) ) or were you replacing the original ( [L254](https://github.com/ilastik/lazyflow/pull/164/files#diff-f2aeda246096a7282ca7ee818a962302L254) )?

stuarteberg: >were you copying the data back into the original array ( L291 ) or were you replacing the original ( L254 )?
My test was based on [L254](https://github.com/ilastik/lazyflow/pull/164/files#diff-f2aeda246096a7282ca7ee818a962302L254).  (Copy/convert-dtype in one step.)

stuarteberg: Thanks for the above explanation.  I think the approach of making operators "mask-aware" on a case-by-case basis for now is fine.  It makes sense to make sure the cache operators are all mask-aware.

lazyflow/stype.py
```diff
@@ -181,7 +191,19 @@ def isConfigured(self):
...
+        # ( https://github.com/numpy/numpy/issues/5558 ).
+        # So, this must be used in the interim.
+        if isinstance(dst, numpy.ma.masked_array):
+            src_val = src[...]
+            dst.data[...] = numpy.ma.getdata(src_val)
+            dst.mask[...] = numpy.ma.getmaskarray(src_val)
+            if isinstance(src_val, numpy.ma.masked_array):
+                dst.fill_value = src_val.fill_value
+        elif isinstance(dst, (list, tuple)) or isinstance(src, (list, tuple)):
+            dst[:] = src[:]
```
stuarteberg: In general, it is better to use `isinstance(dst, collections.Iterable)` for cases like this rather than explicitly listing all the iterable types you can think of...

jakirkham: Sounds good. Though as we need `__getitem__` and `__setitem__` shouldn't we go for something more specific than `collections.Iterable`? Perhaps `collections.MutableSequence`?

lazyflow/operators/opCompressedUserLabelArray.py
jakirkham: > Thanks for the above explanation.

Of course.

> I think the approach of making operators "mask-aware" on a case-by-case basis for now is fine.

Cool.

> It makes sense to make sure the cache operators are all mask-aware.

Cool. Do you think there are any other cache operators that I should be including here?

jakirkham: > My test was based on L254. (Copy/convert-dtype in one step.)

Thanks for clarifying.

jakirkham: Just as a follow-up, I looked into `numpy.where`. This appears to be implemented in C, which would make it a bit tricky to add a fix for `MaskedArray`s as they are implemented in Python. One way to address this, might be to add a Python shim function in place of `numpy.where`, which then calls the C implementation of `numpy.where` or `numpy.ma.where` as needed. This would probably be the easiest way to fix this behavior.

lazyflow/stype.py
jakirkham: Fixed.

burgerdev:
I am not sure if I understand the use case correctly: is the target to be able to use masked arrays in arbitrary lazyflow pipelines? If so, there are vigra functions which are definitely neither aware of nor meaningfully defined for masked arrays and would have to be replaced (e.g. `analysis.labelMultiArray`).

jakirkham:
I think there are two questions that you are asking, which I have listed below. If I am wrong, feel free to correct me. What follows is a short response to the questions and then a bit more detail about the mechanism added if you are interested.

1. What is the intent of this pull request?
2. What affect will this have on current pipelines?
&nbsp;

&nbsp;
As to the first question, the intent of this pull request is to provide a solution to this issue ( https://github.com/ilastik/ilastik/issues/1076 ). A good deal of work has already been done to address this. This particular pull request is a follow up to the previous one for `lazyflow` ( https://github.com/ilastik/lazyflow/pull/163 ). This particular pull request focuses mainly on ensuring that cache operators can cache a masked array if provided, but preserve the normal behavior for regular arrays.

As to the second question, all pipelines will behave the same. In other words, they will continue to get regular arrays and return regular arrays. Having masked arrays as the default sounds dangerous. Our method used in solving this issue was to ensure that the standard behavior from before is the default.
&nbsp;

&nbsp;
The long answer is as follows. A key, `has_mask`, has been added to `MetaDict` as it is uninitialized it defaults to `None`. Lazyflow will not allocate a masked array unless `has_mask` is `True` ( https://github.com/ilastik/lazyflow/commit/4cb052535159fc41cf0425a094d3f58009de8d09 ). There are three ways `has_mask` could become `True`.

1. A user (i.e. developer) changes `has_mask` to `True` for a slot.
2. The input slot is connected to an output slot, whose value for `has_mask` is `True`.
3. A masked array is provided as input to the operator via `setValue`, (added in this pull request) ( https://github.com/jakirkham/lazyflow/commit/227e9c21887b4f2777724c539c4c3ec7aa978eb9 ).

When any one of these three is done, `lazyflow` will make sure that a masked array is used to store the value of the slot. However, this does not guarantee the output slot will return a masked array. To ensure a masked array is returned, the preferred method is the standard method using `assignFrom` to set the meta data of the output slot with the meta data from the input slot. If additional changes to the output meta data are required, they can be done after this step. Alternatively, `has_mask` on the output slot could be set explicitly, but this is frowned upon as other pieces of metadata could have been lost. When this is done, a masked array will be used for the result provided to the `execute` method.
&nbsp;

&nbsp;
I hope this explanation was helpful. If you still have questions, I'd be more than happy to answer them. To make the discussion easier for others to follow, I would really appreciate if general questions get posted to the aforementioned issue ( https://github.com/ilastik/ilastik/issues/1076 ). Of course, if you have comments about specific details, code, etc. feel free to put them where you think it is relevant.

jakirkham:
Added an operator to take a masked array and return a regular array with all masked values filled with the value provided in `InputFillValue`.

lazyflow/stype.py
```diff
@@ -100,13 +101,21 @@ def allocateDestination( self, roi ):
...
+        #     ##storage = storage.view(vigra.VigraArray)
+        #     ##storage.axistags = copy.copy(self.slot.meta.axistags)
         if self.slot.meta.has_mask:
             storage_mask = numpy.zeros(storage.shape, dtype=bool)
-            storage = numpy.ma.masked_array(storage, mask=storage_mask, shrink=False)
-        # if axistags is True:
-        #     storage = vigra.VigraArray(storage, storage.dtype, axistags = copy.copy(s))elf.axistags))
-        #     #storage = storage.view(vigra.VigraArray)
-        #     #storage.axistags = copy.copy(self.axistags)
+            storage_fill_value = None
```
jakirkham: @stuarteberg I would be interested in your feedback here. I have chosen a standard different from NumPy for a few reasons. Of course, these are only the defaults, so the user (i.e. developer) can change them anyways.

1. NumPy picks values that don't make sense for certain types ( http://docs.scipy.org/doc/numpy/reference/maskedarray.baseclass.html#numpy.ma.MaskedArray.fill_value ). In particular, note the value for `int` is used for all integers (i.e. `uint8`), the value for `float` is out of range for `float16` and smaller.
2. It is nice to be able to have the same effect of having masked values even when they aren't there (viz. `nan`). This allows `numpy.isnan` to get back the mask or comparing to `0` in the integer case.

jakirkham:
Rebased on current master to try and add assertions of the form below to all operators that may not support masks at present (i.e. have not been tested).

    assert not self.SomeInput.meta.has_mask, "The operator, \"%s\", is being setup to recieve a masked array as input to slot, \"%s\". This is currently not supported." % (self.name, self.SomeInput.name)

jakirkham:
Ran into an error when running tests after rebase and added a issue related to it ( https://github.com/ilastik/lazyflow/issues/172 ).

jakirkham:
Rebased on fix to issue ( https://github.com/ilastik/lazyflow/issues/172 ).

jakirkham:
@stuarteberg in response to your previous comment ( https://github.com/ilastik/lazyflow/pull/163#issuecomment-73310479 ), [`OpMaskedSelect`]( https://github.com/ilastik/lazyflow/blob/542e6668fd9e6548cd097ebaf47e65fe5d9aba7b/lazyflow/operators/opMaskedSelect.py ) could now be rewritten as a combination of [`OpMaskArray`]( https://github.com/jakirkham/lazyflow/blob/masked_arrays_2/lazyflow/operators/opMaskArray.py ) and [`OpFillMaskArray`]( https://github.com/jakirkham/lazyflow/blob/masked_arrays_2/lazyflow/operators/opFillMaskArray.py ). As for [`OpMaskedWatershed`]( https://github.com/ilastik/lazyflow/commit/ef062a70a34b9c711a287b75baf6b3b5ec72e0ed ), we could make this mask aware by modifying the operator slightly or we could use `OpMaskArray` on the inputs to reduce down to a single masked array, whose data and mask could be used separately.

jakirkham:
Added another helper operator for splitting up the content of a masked array into its components. Also, improved the fill operator to allow optionally using the the masked arrays supplied `fill_value` parameter.

jakirkham:
@stuarteberg do you have any thoughts on multilevel slot support?

jakirkham:
Added the keyword argument `allows_mask` to the class `Slot`. This value defaults to `False`. As a result, when a masked array is provided as input it will result in an `AssertionError`. To avoid this, `allows_mask` must be set to `True` on all `Slot`s that will be provided or return a masked array.

jakirkham:
Added some more tests for `allow_mask`. These verify that connections between something that `has_mask` and one that does not enable `allow_mask` will result in an `AssertionError`.

jakirkham:
Rebased. Also, added masked array support (with tests) for `OpArrayPiperWithAccessCount`.

lazyflow/operators/opCompressedUserLabelArray.py
jakirkham: It's worth noting that previously there was an inconsistency in the implementation of the `numpy.where` and `numpy.ma.where` as is mentioned here ( https://github.com/numpy/numpy/issues/5679 ). A fix has been made and now has been merged into `numpy`'s master branch and should show up in version 1.10.

jakirkham:
As previously mentioned, workarounds for this bug ( https://github.com/numpy/numpy/pull/5580 ) in NumPy were included. A fix is planned for NumPy 1.12.0 and is presented in this PR ( https://github.com/numpy/numpy/pull/5580 ).

stuarteberg:
Thanks for the heads-up.

jakirkham:
Sorry, it is taking awhile. It turned out the old behavior of having data and masks get out of sync was intentional and documented as such. So, it required a deprecation ( https://github.com/numpy/numpy/pull/7020 ).

PR #166, opened 2015-02-13
tests/testOpFormattedDataExport.py: Fix unsigned rollover issue

jakirkham:
This is to address ilastik#165. Test fails sometimes due to unsigned rollover issue (i.e. assuming `numpy.uint8` as our type we have `1-2=255`). Here we simply promote to the next largest signed type (i.e. `numpy.int16`). This way, we can still fit all possible values we had before. Also, we can now have `-1`, which puts `numpy.abs` to use.

### DIFF
--- a/tests/testOpFormattedDataExport.py
+++ b/tests/testOpFormattedDataExport.py
@@ -88,7 +88,11 @@ class TestOpFormattedDataExport(object):
 
         # Due to rounding errors, the actual result and the expected result may differ by 1
         #  e.g. if the original pixel value was 32.99999999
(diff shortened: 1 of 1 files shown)


jakirkham:
@stuarteberg Please let me know your thoughts. This should, hopefully, be easy.

stuarteberg:
Nice catch.

jakirkham:
Thanks. So, I'm confused. Why aren't we merging this?

stuarteberg:
Oops, I hit the wrong button.

jakirkham:
No worries. Thanks.

PR #176, opened 2015-03-11
Fix failing unittests

burgerdev:
There are two failing unittests in HEAD.


stuarteberg:
Jeez, that's sloppy of me.  Thanks for cleaning up after me.

PR #181, opened 2015-03-31
Drtile replacement

burgerdev:
I wrote a pure Python replacement for drtile, which tiles a boolean array into rectangles. For a simple example, see the test file. The test file can also be run standalone to get a benchmark of C++ vs. Python. I have the results from my netbook [here](https://gist.github.com/burgerdev/f3df7e9a50dcc5dc2488). What I see from this is that the performance is

  * worst case half as good as C++
  * for 'normal' tilings still in the order of 10ms

I would consider this alone good enough for replacing the C++ library, but it gets even better. It is strictly beyond me what [the C++ source](https://github.com/ilastik/lazyflow/blob/master/lazyflow/drtile/drtile.cpp#L143) is actually doing, but I figured out that it's not tiling very nice. In fact, an array like this

```
[[1 1 1 1 1]
 [1 1 1 1 1]
 [1 1 0 0 1]
 [1 1 0 0 1]
 [0 1 1 1 1]]
```
is tiled as
```
[[1 2 3 5 7]
 [1 2 3 5 7]
 [1 2 0 0 7]
 [1 2 0 0 7]
 [0 2 4 6 7]]
```
where an optimal solution would look more like
```
[[1 1 2 2 4]
 [1 1 2 2 4]
 [1 1 0 0 4]
 [1 1 0 0 4]
 [0 3 3 3 4]]
```
However, finding the optimal solution  is a packing problem and NP hard if I am not mistaken. But we can do better than just computing vertical tilings. In fact, the python implementation does extend tiles in all dimensions if possible and comes up with this

```
[[1 2 4 4 5]
 [1 2 4 4 5]
 [1 2 0 0 5]
 [1 2 0 0 5]
 [0 2 3 3 5]]
```

which is not optimal but still two requests better than our current version.

### DIFF
--- a/lazyflow/drtile2/__init__.py
+++ b/lazyflow/drtile2/__init__.py
@@ -1,0 +1,23 @@
+###############################################################################
+#   lazyflow: data flow based lazy parallel computation framework
+#
+#       Copyright (C) 2011-2014, the ilastik developers
+#                                <team@ilastik.org>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the Lesser GNU General Public License
+# as published by the Free Software Foundation; either version 2.1
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# See the files LICENSE.lgpl2 and LICENSE.lgpl3 for full text of the
+# GNU Lesser General Public License version 2.1 and 3 respectively.
+# This information is also available on the ilastik web site at:
+#          http://ilastik.org/license/
+###############################################################################
+
+from drtile import drtile
--- a/lazyflow/drtile2/drtile.py
+++ b/lazyflow/drtile2/drtile.py
@@ -36,10 +36,15 @@ def _drtile_old(bool_array):
 
 def _drtile_new(bool_array):
     """
-    tile the true parts of a bool arrray
+    tile the non-zero parts of a bool array into rectangles
  
-    Returns an iterable that contains rois as concatenated start and
+    Returns an iterator that contains rois as concatenated start and
     stop vectors (for a 3-dim array, roi.shape == 6).
+
+    Note: If you want to keep the tiles for more than one iteration, use
+    this function as
+
+    >>> tile_rois = list(drtile(bool_array))
     """
     x = bool_array
 
@@ -59,7 +64,7 @@ def _drtile_new(bool_array):
             for k in current)
 
         # finalize tilings that are not in the current hyperplane
-        final = ifilter(lambda k: k not in current, last_dict)
+        final = ifilter(lambda k: k not in current_dict, last_dict)
         expanded = imap(lambda k: _expand_roi(k, last_dict[k], i), final)
         rois.extend(expanded)
 
@@ -67,7 +72,7 @@ def _drtile_new(bool_array):
 
     # add the remaining tiles
     expanded = imap(lambda k: _expand_roi(k, last_dict[k], x.shape[-1]),
-                    last)
+                    last_dict)
     rois.extend(expanded)
     return rois
 
--- a/lazyflow/drtile2/drtile.py
+++ b/lazyflow/drtile2/drtile.py
@@ -1,0 +1,105 @@
+###############################################################################
+#   lazyflow: data flow based lazy parallel computation framework
+#
+#       Copyright (C) 2011-2014, the ilastik developers
+#                                <team@ilastik.org>
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the Lesser GNU General Public License
+# as published by the Free Software Foundation; either version 2.1
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# See the files LICENSE.lgpl2 and LICENSE.lgpl3 for full text of the
+# GNU Lesser General Public License version 2.1 and 3 respectively.
+# This information is also available on the ilastik web site at:
+#          http://ilastik.org/license/
+###############################################################################
+
+from itertools import imap, ifilter
+
+import numpy as np
+import vigra
+
+from lazyflow.drtile import drtile as drtilecpp
+from lazyflow.utility.fastWhere import fastWhere
+
+
+def _drtile_old(bool_array):
+    tileWeights = fastWhere(bool_array, 1, 128**3, np.uint32)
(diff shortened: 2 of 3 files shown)


burgerdev:
I see this more as a proof of concept to talk about rather than an actual pull request to merge. The final implementation could be tied in much better with `OpArrayCache`, the only place where `drtile` is used throughout lazyflow.

stuarteberg:
Just to make sure we're on the same page: drtile still exists in the code base, but it is entirely an optional dependency now.  (If you don't compile it, it doesn't get used.)  I was thinking about just dropping it entirely.

My reasons are the following (all of which apply to the old implementation of drtile.  I assume they apply to this proposed implementation, too)

- It isn't clear to me that the problem drtile attempts to solve is a common scenario in general ilastik usage.  For example, how often does ilastik need to request a big area from a cache that is already partially clean in the middle, but not on the borders?
- And in such cases, how much slower is it to send out lots of small requests than it is to send fewer big requests?  We never tested to see if this improved performance at all (either throughput or latency).  It may be worse, for all I know.
- It can make debugging a workflow a little more complicated, because the requests will be merged into shapes  you might not have expected.  Furthermore, the requested shape won't necessarily be the same from one run to the next.
- It makes it difficult to estimate the RAM required for a particular workflow.  As it turns out, caches serve two orthogonal purposes: they break up big requests into small requests, and they also cache the data.  But drtile actually "undoes" the first effect.  Sometimes that is exactly the opposite of what a workflow author wanted.

(Note to self: We should consider making these two orthogonal purposes explicitly separated, perhaps by implementing `OpArrayCache` as a combination of `OpUnblockedArrayCache` followed by some new `OpDivideRequests` operator.)

Having said all that, I do like the fact that your solution is Python-only, and that you benchmarked it against the C++ alternative.  If we decide to keep drtile, I'm in favor of this implementation.

burgerdev:
> For example, how often does ilastik need to request a big area from a cache that is already partially clean in the middle, but not on the borders?

It happens when zooming out or panning: I assume it's the phenomenon when your freshly computed segmentation trickles in tile-by-tile after clicking the zoom-out button,

> And in such cases, how much slower is it to send out lots of small requests than it is to send fewer big requests?

I ran a quick [benchmark](https://gist.github.com/burgerdev/864ae1bfb5d6a38520d0) and found a significant overhead in the order of seconds just from those small requests (using only array piper and cache with a megapixel image).

> It can make debugging a workflow a little more complicated, because the requests will be merged into shapes you might not have expected.
> [...]
> It makes it difficult to estimate the RAM required for a particular workflow. As it turns out, caches serve two orthogonal purposes: they break up big requests into small requests, and they also cache the data.

I am not sure if I understand this argument correctly. Where do workflow developers expect their requests to come in rois of 64x64x1? I have not seen an example of that in PC or OC, but I don't know about carving or tracking. Also, most non-trivial operators do need to change the roi in one way or another (halos, ...), so expecting a particular request roi pattern might be unwise.

The reason for drtile was probably something like "requests are bad and produce to much overhead", as far as I understand it. And I think this is still a valid assumption.

Related: drtile should perhaps be decoupled from lazyflow, since it does not add much value in a headless setting but mostly accomodates volumina.

stuarteberg:
>Where do workflow developers expect their requests to come in rois of 64x64x1?

Never.  You're right: without drtile, the value of `OpArrayCache.DefaultBlockSize` is an unfortunate choice.  But in practice, no operators actually use this default.  Every `OpArrayCache` in ilastik or lazyflow uses an explicit blockshape provided by the workflow author.

>I ran a quick benchmark and found a significant overhead in the order of seconds just from those small requests (using only array piper and cache with a megapixel image).

Nice work!  But I think the benchmark needs to be tuned.

- drtile is more likely to shine if your caches with lots of small blocks.  In this benchmark, the 'default' (but never used) block size of (64,64) is used.  In real ilastik workflows, the blocks are more like (256,256) -- i.e. 16 times fewer requests.
- In ilastik, the latency of a request comes from multiple sources:
 - volumina request latency (before lazyflow, e.g. converting a request for image data to a request for 2d slice data to a request for 5D data)
 - compositing several layers into a single QImage tile.
 - performing the actual computation (e.g. reading data from disk, or computing a feature)
 - request latencies

 Drtile can't really help with the first three (well, it helps a little with reading data because it can reduce the total number of halo pixels), so it mainly improves the last item.  But the benchmark above only tests the last one.  Even if drtile helps with this last source of latency, what percentage of the TOTAL latency does that amount to?  Is it worth adding complexity to the cache to accommodate it?  I honestly do not know the answer, which isn't good.  We don't have a great way to measure the relative contributions of each of the above factors.  Do you have any ideas?  As at least a quick start, your benchmark could do some "real" work instead of returning a precomputed array.

I don't mean to be a nay-sayer.  I'm glad you're looking into this issue.

burgerdev:
I've got some [log messages](https://gist.github.com/burgerdev/4fd6e60db5a49889a0b2) from a pixel classification session that indicate typical caches in the interior of the volume to be of shape 256x256 with inner shape 128x128 or 64x64. For completeness, I ran the benchmark with block numbers from 1 to 16, resulting in the following plot.

drtile

> Even if drtile helps with this last source of latency, what percentage of the TOTAL latency does that amount to? Is it worth adding complexity to the cache to accommodate it?

I am not sure about this either, but my intuition tells me that we should take what we can possibly get in terms of latency reduction. On the other hand, I am all for simplifying those trivial-on-paper cache operators. 

My general suspicion is that these small time differences will add up because of the large amount of requests that are generated. Later today, I will run some ilastik sessions with all three methods to see if there are visible changes in latency with one or the other.

burgerdev:
Check this out: [changing one line of source code causes major slowdown](https://vimeo.com/123870759) (video might not be ready yet, due on Apr 1 2015, UTC 20:00). The video shows me using a pre-trained object classification project, traversing time slices and clicking on the zoom-to-fit button. Waiting times are nearly twice as long without the c++ tiling.

stuarteberg:
That is quite interesting.  One update: I just added a [commit][1] on the ilastik master branch that causes all blocked array caches to use the same shape for inner and outer block shapes.  This will bring the requests back down by a factor of 4.  If you end up doing more benchmarking, can you include that change?

[1]: https://github.com/ilastik/ilastik/commit/eed996a141638cf723dab9a97171822193f700d9

burgerdev:
I re-ran the benchmark including the changes from ilastik/ilastik@eed996a141638 and observed no significant differences between the three methods, and there was an improvement by something about 30% compared to before ilastik/ilastik@eed996a141638. 

What I am wondering at now: Do we need `OpArrayCache` at all when each `OpBlockedArrayCache` contains just one `OpArrayCache`?

stuarteberg:
>What I am wondering at now: Do we need OpArrayCache at all when each OpBlockedArrayCache contains just one OpArrayCache?

Yeah, I agree.  I think `OpBlockedArrayCache` could have a much simpler (and probably faster) implementation. And `OpArrayCache` could probably be deprecated, maybe just used for testing.

As I mentioned above in my "note to self", one possible implementation of `OpBlockedArrayCache` is to simply chain together an `OpUnblockedArrayCache` with a new operator named `OpDivideRequests`.

burgerdev:
So I guess we will opt for abandoning OpArrayCache and will have no need for drtile (or a replacement) in the future. Case closed.

PR #182, opened 2015-04-07
Refactor blocked array cache

stuarteberg:
Here's a new implementation of `OpBlockedArrayCache` (tentatively named `OpRefactoredBlockedArrayCache`). That separates the tasks of "caching", "fixing", and "splitting" into separate operators:

    # Input ---------> opCacheFixer -> opUnblockedArrayCache -> opSplitRequestsBlockwise -> Output
    #                 /                                        /
    # fixAtCurrent --                                         /
    #                                                        /
    # BlockShape --------------------------------------------

Each component operator is simpler to understand.

TODO:

- [x] Deal with `FIXME` mentioned in `OpRefactoredBlockedArrayCache`: this implementation needs RAM even for the zero blocks while `fixAtCurrent=True`
- [x] Write unit tests for `OpCacheFixer` and `OpSplitRequestsBlockwise`
- [x] Incorporate a `Cache` mixin after PR #160 is merged.
- [ ] Some slots have unconventional names (`outerblockShape`, etc.), for compatibility with the old `OpBlockedArrayCache`.  If we decide to switch to this new implementation, we can rename those slots with standard conventions.  Also, we can get ride of the `innerBlockShape` slot.


stuarteberg:
@burgerdev, can you take a look at this when you get a chance?

burgerdev:
Sure.

burgerdev:
Regarding `OpCacheFixer`: I can imagine situations were we wouldn't want to have one big rectangular region that encompasses all dirty regions ever encountered, but fewer smaller ones. Should we write two operators for this? I am not quite sure what would be correct here.

burgerdev:
I wrote the unit tests and merged master. The cache mixin is already available through `OpUnblockedArrayCache`, that should be sufficient.

Imho, there is no easy solution that both separates fixing from caching and does not cache 0s.

burgerdev:
Turns out that `OpSlicedBlockedArrayCache` needs `OpBlockedArrayCache` to implement the `Cache` interface, added it to the PR.

stuarteberg:
>Regarding `OpCacheFixer`: I can imagine situations were we wouldn't want to have one big rectangular region that encompasses all dirty regions ever encountered, but fewer smaller ones.

I agree there could be such situations.  Do you think any of them are common scenarios?  I don't want to send out lots of small dirty notifications (e.g. one for every block) because that would be very slow.

>I wrote the unit tests and merged master.

Thanks! I merged your PR into this branch.

>Imho, there is no easy solution that both separates fixing from caching and does not cache 0s.

I was inclined to agree, but now I think I solved it with a quick hack. See my latest commit.

burgerdev:
> Do you think any of them are common scenarios?

What happens when user-drawn labels change in distant regions? Is the user label operator still a blocked cache (or has it ever been one)?

> I think I solved it with a quick hack

I like that. Did you intentionally not go for a new meta tag? I figure it would be easier to reuse if no custom array classes were involved..

stuarteberg:
>Did you intentionally not go for a new meta tag? I figure it would be easier to reuse if no custom array classes were involved..

Hmm... I thought I had a reason, but I can't think of it now.  You're right: using the slot metadata would be more idiomatic.  I'll change it.

stuarteberg:
>Is the user label operator still a blocked cache (or has it ever been one)?

No.  User-drawn labels are stored with `OpCompressedUserLabelArray`, which is not a `Cache` under the new scheme.

PR #186, opened 2015-04-13
[Ready] Split RAM into cache memory and computation memory

### LINKED ISSUE #185: Cache cleanup strategy broken
The current strategy for cache cleanup 

  1. Check current process memory usage `m`
  2. If `m` > `0.85*LAZYFLOW_TOTAL_RAM` throw out a cache
  3. repeat

does not work well when batch processing or exporting. When these operations occupy large areas of available memory themselves, all caches are thrown out and - worst case - there is an infinite request-throwout-request_again loop. I think that the memory consumption computation by `meta.ram_usage_per_requested_pixel` has its part in this[1], too, but we probably can't rely on that to work perfectly. 

Suggested fix: Target a **percentage of available memory for caches**. E.g.:

  1. Fix max memory `M` for caches (`.5*LAZYFLOW_TOTAL_RAM` ?)
  2. Check current cache memory usage `m`
  3. If `m>M` throw out a cache
  4. repeat from 2

[1]: I am using Pixel Classification + Object Prediction "from scratch". When exporting via the export applet, RAM used by ilastik goes up to 3 times more than allowed by `LAZYFLOW_TOTAL_RAM_MB`

burgerdev:
fixes #185


burgerdev:
The default memory used is limited to 3/4 of the available memory to keep ilastik from swapping other apps out by default. This is probably a bad configuration for large machines, but I guess those will have a custom `.ilastikrc` entry anyways.

stuarteberg:
>The default memory used is limited to 3/4 of the available memory to keep ilastik from swapping other apps out.

FWIW, when this subject has been raised in the past, we decided that we *want* to swap other apps out.  If the user has 100 tabs open her web browser and is also using Blender and Photoshop, ilastik won't have much RAM to work with if we don't let it force the other apps into the swap space.  We assume that the user wants ilastik to take up as much RAM as possible, and if she wants to limit ilastik to less than that, then she can edit `.ilastikrc`.  Does that sound reasonable to you?

burgerdev:
Sounds reasonable, but I guess we have to keep in mind that the default settings will always cause swapping in batch mode or in the export applet when we set the available ram to 100%.

stuarteberg:
>the default settings will always cause swapping in batch mode or in the export applet when we set the available ram to 100%. Perhaps we can introduce a safety factor somewhere else?

OK.  Perhaps we should make the default "available" RAM only 80% of physical RAM?

burgerdev:
> OK. Perhaps we should make the default "available" RAM only 80% of physical RAM?

@stuarteberg I was just thinking about our multiple target platforms (laptop, office machine, cluster): How about a default limit of

```python
max(_physically_available_ram * 0.8, _physically_available_ram - 1024**3)
```

On a large machine we would still use almost all of the available RAM, on a laptop we would use 80%.

stuarteberg:
>On a large machine we would still use almost all of the available RAM, on a laptop we would use 80%.

I'm not sure I see the benefit.  The amount of RAM we need to reserve for the OS ought to be fairly constant across machines, as long as they're running the same OS.  Maybe instead of using 80% in all cases, we should use `_physically_available_ram - 1024**3` in all cases.

Either way, I think we'll have an easier time debugging this stuff and supporting users if the formula is simple and doesn't change too much from one machine to another.  Let's just pick something that strikes you as reasonable for now, and change it later if we run into problems.

PR #191, opened 2015-06-09
Added Vigra RF Adaptive Mask Pixelwise Classifier

JaimeIvanCervantes:
Added classes VigraRfAdaptiveMaskPixelwiseClassifierFactory and VigraRfAdaptiveMaskPixelwiseClassifier. By adaptive mask I am referring to a mask that is calculated every certain number of frames to identify ROIs around relevant objects and discard the rest of the pixels as background.

### DIFF
--- a/lazyflow/classifiers/vigraRfAdaptiveMaskPixelwiseClassifier.py
+++ b/lazyflow/classifiers/vigraRfAdaptiveMaskPixelwiseClassifier.py
@@ -1,0 +1,198 @@
+import os
+import tempfile
+import cPickle as pickle
+
+import numpy
+import scipy.ndimage
+import vigra
+import h5py
+
+from lazyflowClassifier import LazyflowPixelwiseClassifierABC, LazyflowPixelwiseClassifierFactoryABC
+from lazyflow.utility import Timer
+
+import logging
+logger = logging.getLogger(__name__)
+
+class VigraRfAdaptiveMaskPixelwiseClassifierFactory(LazyflowPixelwiseClassifierFactoryABC):
+    """
+    An implementation of LazyflowPixelwiseClassifierFactoryABC using a vigra RandomForest.
+    This exists for testing purposes only. (it is normally better to use the vector-wise 
+    classifier so lazyflow can cache the feature matrices).
+    This implementation is simple and un-optimized.
+    """
+    VERSION = 1 # This is used to determine compatibility of pickled classifier factories.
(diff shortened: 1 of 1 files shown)


lazyflow/classifiers/vigraRfAdaptiveMaskPixelwiseClassifier.py
```diff
@@ -0,0 +1,198 @@
...
+                and self._kwargs == other._kwargs )
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+assert issubclass( VigraRfAdaptiveMaskPixelwiseClassifierFactory, LazyflowPixelwiseClassifierFactoryABC )
+
+class VigraRfAdaptiveMaskPixelwiseClassifier(LazyflowPixelwiseClassifierABC):    
+    """
+    Adapt the vigra RandomForest class to the interface lazyflow expects.
+    """
```
stuarteberg: This class needs a better docstring.  Since the idea here is to put this work and pick it up later, we'll need help remembering what this class does and what the next steps to take are.  Regarding the next steps, mention some TODO items, such as parallelization, profile/optimize the masking step (is there a faster option than scipy's `binary_dilation()` function?), and also mention the need to think about issues at the image borders.  (The current implementation works if you're predicting the entire frame, but for spatially-divided blocks, objects that "walk in" from the side will be missed.)

lazyflow/classifiers/vigraRfAdaptiveMaskPixelwiseClassifier.py
```diff
@@ -0,0 +1,198 @@
...
+        
+        # Allocate memory for probability volume and mask    
+        prob_vol = numpy.zeros((X.shape[:-1]+(len(self._known_labels),) ))           
+        mask = numpy.ones(numpy.prod(X.shape[1:-1]))
+        
+        frm_cnt = 0
+          
+        for X_t in X:    
+            if frm_cnt % FRAME_SPAN == 0 :
+                mask = numpy.ones(numpy.prod(X.shape[1:-1]))
```
stuarteberg: When using `numpy.zeros()` or `numpy.ones()`, always specify an explicit dtype.  The default dtype is `float64`, which is larger than necessary here, and therefore slower.  Need to fix this line and a few others, below.

JaimeIvanCervantes:
- Added `dtype` to numpy array memory pre-allocation. 
- Added TODO comments and other comments that might be helpful in the future.

PR #193, opened 2015-07-04
fixes #192

### LINKED ISSUE #192: Late initialization of Request._lock leads to test failure
See the background thread error in [this CI run](https://circleci.com/gh/ilastik/lazyflow/37). Additionally, we should make sure that we don't shoot ourselves in the foot by randomness in `testValueProviders:TestOpValueCache.test_cancel`.

burgerdev:
fixes #192

### DIFF
--- a/tests/testValueProviders.py
+++ b/tests/testValueProviders.py
@@ -205,6 +205,7 @@ class TestOpValueCache(object):
         threads, even if one thread cancels its request.
         The OpValueCache must handle Request.InvalidRequestException errors correctly. 
         """
+        n = 20
         graph = lazyflow.graph.Graph()
         opCompute = TestOpValueCache.OpSlowComputation(graph=graph)
         opCache = OpValueCache(graph=graph)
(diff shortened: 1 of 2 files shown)


stuarteberg:
Nice catch.

tests/testValueProviders.py
```diff
@@ -235,8 +243,13 @@ def checkOutput():
         for t in threads:
             t.join()
 
+
+        numpy.testing.assert_equal(opCompute.executionCount, 1)
```
stuarteberg: As the CI testing has just revealed, this assertion seems occasionally fail.  I'm not sure that it's correct.  In the current implementation of `OpValueCache`, it is possible for the value to be requested more than once if multiple requests are created and cancelled very quickly.  I don't think that's a problem, as long as the requests that *weren't* cancelled eventually get a result.  I'm going to remove this assert.  If you think that's wrong, let me know.

tests/testValueProviders.py
```diff
@@ -235,8 +243,13 @@ def checkOutput():
...
+        assert opCache._dirty == False
```
stuarteberg: And this assert also fails occasionally, because of line 232.  I think the assert is valid if we request output from the cache one last time.  Will push a fix in a minute.

tests/testValueProviders.py
burgerdev: You're right.

tests/testValueProviders.py
burgerdev: I am not sure how I could miss that line 232,. Does it even make sense?

stuarteberg: The reason for line 232 is that this test wants to force the cache to request the data from the upstream source, not to serve it out of memory.  I think it makes sense, and I think the test is fixed now.

PR #194, opened 2015-07-10
Fixed error in the function create_and_train() of parallelVigraRfLazyflowClassifier.py

JaimeIvanCervantes:
there is an error that can be recreated when the number of trees is less than the number of processors. In the function create_and_train() of parallelVigraRfLazyflowClassifier.py, the list tree_counts has trailing zeros when _num_trees is less than _num_forests.


stuarteberg:
It's nice if PRs can have a clean git history.  Can you collapse these changes into 1 commit and re-push?  There are a couple ways to do this, but these commands ought to do the trick:

```
git reset HEAD^
git add lazyflow/classifiers/parallelVigraRfLazyflowClassifier.py
git commit --amend --reuse-message HEAD
git push -f JaimeIvanCervantes master # <-- Or whatever your 'remote' is named instead of JaimeIvanCervantes
```

burgerdev:
While you are at it: rebasing your changes onto the current master should make CI tests pass.

PR #201, opened 2015-09-10
Added features for parameter-optimization to the Paralllel Vigra RF Classifier Factory

JaimeIvanCervantes:
Added the following features to the Parallel Vigra RF Classifier Factory:
* Public setter functions for the number of trees, the path to the variable importance table, and the proportion of labels.
* Updated the averaging of the feature importance tables (from forests running in parallel) to a weighted average that takes into account the number of trees per thread.

These new features were added to enable the parameter-optimization tests (features and number of trees).

### DIFF
--- a/lazyflow/classifiers/parallelVigraRfLazyflowClassifier.py
+++ b/lazyflow/classifiers/parallelVigraRfLazyflowClassifier.py
@@ -18,7 +18,7 @@ logger = logging.getLogger(__name__)
 
 
 class ParallelVigraRfLazyflowClassifierFactory(LazyflowVectorwiseClassifierFactoryABC):
-    VERSION = 1 # This is used to determine compatibility of pickled classifier factories.
+    VERSION = 2 # This is used to determine compatibility of pickled classifier factories.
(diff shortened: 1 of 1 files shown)


lazyflow/classifiers/parallelVigraRfLazyflowClassifier.py
```diff
@@ -14,11 +16,12 @@
 import logging
 logger = logging.getLogger(__name__)
 
+
 class ParallelVigraRfLazyflowClassifierFactory(LazyflowVectorwiseClassifierFactoryABC):
     VERSION = 1 # This is used to determine compatibility of pickled classifier factories.
                 # You must bump this if any instance members are added/removed/renamed.
```
stuarteberg: You'll have to increment this `VERSION` constant.  (I forgot to consider this previously.)

Background: `ClassifierFactory` instances are saved to the project file using the python `pickle` module.  If you add or remove instance members, then the old pickle representation is out-of-sync with the new class.  Bump the version so ilastik knows not to re-use saved classifier factory instances from old project files.

PR #208, opened 2016-01-29
Learning - using crops for selecting training regions

ilastikdev:
Required for crop selection and training applets in the new Structured Learning Tracking workflow.


stuarteberg:
As far as I can tell, the class you are attempting to merge here is just a copy of the existing `OpCompressedUserLabelArray` class, with the exception of some slot/variable name changes.  Is there a difference I'm missing?

If not, I think it would be better to delete this class and just use the existing one.  If there is some subtle difference that I didn't notice, then the operator should be implemented in such a way that avoids code duplication with the original.

ilastikdev:
ilastik/lazyflow learning is ready to be pulled into 
ilastik/lazyflow master

stuarteberg:
Your most recent commits effectively revert the previous changes in this branch, leaving this PR empty.  (That's why github's "Files changed" view above has 0 files listed.)

Instead of merging a string of commits that do nothing in the end, is it okay to just close this PR without merging it?

ilastikdev:
of course

PR #212, opened 2016-04-26
opCompressedCache: Set the ideal_blockshape to the shape of a single frame

JaimeIvanCervantes:
Set the `ideal_blockshape` of the compressed cache to the shape of a single frame instead of the shape of the HDF5 chunk.


stuarteberg:
Oops, now the CI server is complaining.  Looks like we need to update the test suite.  Can you investigate?

https://circleci.com/gh/ilastik/lazyflow/196

PR #221, opened 2016-11-03
Hdf5 Cache

holstgr-kaust:
This patch is based on h5-cache, but updated to sync with master.  The primary changes were made to opUnblockedHdf5Cache, making the locking simpler, and making propagate dirty work correctly.  It works with the ilastk/blockwise_carving patch.

### DIFF
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -422,9 +422,7 @@ class Operator(object):
         @functools.wraps(func)
         def wrapper(self, *args, **kwargs):
             with self._condition:
-                # TODO: verify what the correct test should be
-                # while self._executionCount > 0:
-                while self._executionCount > 1:
+                while self._executionCount > 0:
                     self._condition.wait()
                 self._settingUp = True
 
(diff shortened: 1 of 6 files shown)


k-dominik:
in general I can see a `hdf5` (or some other file back-end cache) coming in handy, however I would be interested if there was something specific this PR aimed at... can you maybe remember @akreshuk ?

holstgr-kaust:
> in general I can see a `hdf5` (or some other file back-end cache) coming in handy, however I would be interested if there was something specific this PR aimed at... can you maybe remember @akreshuk ?
This PR provides support for ilastik PR 1365 Blockwise Carving [https://github.com/ilastik/ilastik/pull/1365] and ilstiktools PR 4 Blockwise Carving [https://github.com/ilastik/ilastiktools/pull/4].  There were expensive calculations, done in parallel, that wouldn't all fit in memory; hence the file cache.  Changes were made to original opBlockedHdf5Cache by Stuart to more simply track which blocks were already calculated (as I recall).

If there is interest in reviving these three PRs, I can update them to the latest versions of ilastik / lazyflow / ilastiktools, and provide rational for the changes (it's from over two years ago, so I'll have to review them myself too).

Hope that helps.

PR #224, opened 2017-01-10
1: Baselining for python 3

njase:
This branch is only for the purpose of review request


k-dominik:
With the python 3 switch done, I think we can close this one.

PR #225, opened 2017-03-06
Add npz support

k-dominik:
Very basic support for numpy npz files added. Works for uncompressed and compressed npz archives. However, numpy doesn't offer memory mapping of npz files. So the whole array is loaded.

### DIFF
--- a/lazyflow/operators/ioOperators/opInputDataReader.py
+++ b/lazyflow/operators/ioOperators/opInputDataReader.py
@@ -73,7 +73,7 @@ class OpInputDataReader(Operator):
     videoExts = ['ufmf', 'mmf', 'avi']
     h5Exts = ['h5', 'hdf5', 'ilp']
     klbExts = ['klb']
-    npyExts = ['npy']
+    npyExts = ['npy', 'npz']
     rawExts = ['dat', 'bin', 'raw']
     blockwiseExts = ['json']
     tiledExts = ['json']
--- a/lazyflow/operators/ioOperators/opInputDataReader.py
+++ b/lazyflow/operators/ioOperators/opInputDataReader.py
@@ -29,7 +29,7 @@ from opTiffSequenceReader import OpTiffSequenceReader
 from lazyflow.operators.ioOperators import OpStackLoader, OpBlockwiseFilesetReader, OpRESTfulBlockwiseFilesetReader, \
     OpCachedTiledVolumeReader, OpKlbReader
 from lazyflow.utility.jsonConfig import JsonConfigParser
-from lazyflow.utility.pathHelpers import isUrl
+from lazyflow.utility.pathHelpers import isUrl, PathComponents
 
 from opStreamingUfmfReader import OpStreamingUfmfReader
 from opStreamingMmfReader import OpStreamingMmfReader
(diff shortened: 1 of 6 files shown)


lazyflow/operators/ioOperators/opInputDataReader.py
```diff
@@ -73,7 +73,7 @@ class OpInputDataReader(Operator):
     videoExts = ['ufmf', 'mmf', 'avi']
     h5Exts = ['h5', 'hdf5', 'ilp']
     klbExts = ['klb']
-    npyExts = ['npy']
+    npyExts = ['npy', 'npz']
```
stuarteberg: In addition to this change, you should consider actually setting the `InternalPath` in `OpInputDataReader._attemptOpenAsNpy()`, below.

We can assume the same conventions we use for hdf5.  That is, the `FilePath` value can be parsed as `/path/to/file.npy/internalName`.  We have a little utility class named `PathComponents` defined in `lazyflow.utility` that can help with this. (See `lazyflow/utility/pathHelpers.py`.)  You'll have to make a slight modification to the `PathComponents` source code, since currently it only allows files with an hdf5 extension to have an `internalPath`.

Once you do all that, then ilastik's headless mode will support `.npz` files with user-specified internal paths.  Supporting them in the GUI, too, would require yet more work, and probably isn't worth the trouble right now.

tests/testOpNpyFileReader.py
```diff
@@ -65,6 +70,38 @@ def test_OpNpyFileReader(self):
...
+            # Read the entire file and verify the contents
+            a = npyReader.Output[:].wait()
+            assert a.shape == (10, 11)  # OpNpyReader automatically added a channel axis
+            assert npyReader.Output.meta.dtype == self.testData.dtype
+
+            numpy.testing.assert_almost_equal(a, self.testData)
+        finally:
+            npyReader.cleanUp()
+
+    def test_OpNpyFileReaderNPZnamed(self):
```
k-dominik: Thank you for this hint - the reader indeed wasn't behaving in the expected way.

lazyflow/operators/ioOperators/opNpyFileReader.py
```diff
@@ -25,11 +25,17 @@
...
+import logging
+logger = logging.getLogger(__name__)
+
+
 class OpNpyFileReader(Operator):
     name = "OpNpyFileReader"
     category = "Input"
 
     FileName = InputSlot(stype='filestring')
+    InternalPath = InputSlot(stype='string', value="")
```
k-dominik: With `InternalPath` as an optional InputSlot, the order in which 'FileName' and 'InternalPath' are set becomes important. This can be handled by `OpInputDataReader`. Is this behaviour something we can live with?

stuarteberg: Ah, good point.  Yes, I think it's something we can live with.  To be honest, you've hit upon a more general flaw in the design of lazyflow.  The operator is automatically configured as soon as one of its inputs changes, which isn't always desirable, especially if you intend to change multiple slots in sequence.  In lazyflow's successor, we need to make sure this doesn't happen.

For now, you have to make sure that the operator does something reasonable, or at least gives a clear error message.

stuarteberg:
LGTM!

PR #226, opened 2017-03-13
Added hdf5 sequence reader

k-dominik:
This pull request is related to ilastik issue [1361](https://github.com/ilastik/ilastik/issues/1361).
It comprises the lazyflow parts that enable importing sequences inside hdf5 files.


chaubold:
This is great news! 

I haven't looked into the code, but for our Fiji interoperability this could be helpful, too: does this PR also provide an easy possibility to load several datasets from the same HDF5 file and stack those over z/t?

akreshuk:
I suppose we should wait for the (failing) tests to become passable again? Or is it ready to merge? In either case, good to see it, the users have been waiting for this!

k-dominik:
> does this PR also provide an easy possibility to load several datasets from the same HDF5 file and stack those over z/t?

That's exactly what it does @chaubold! What it doesn't provide though, is the functionality of stacking datasets from different files. Is that something I should look into?

> I suppose we should wait for the (failing) tests to become passable again? 

@akreshuk: Do you reference the commit messages of 9d6530d and c6e70b3, or did you run the lazyflow tests and something failed there? As for the commit messages, those were failing after the specific commits (because only tests were addded). Subsequent commits implement the tested functionality. All lazyflow tests (except the skipped ones) passed on my machine. Maybe I should have edited the commit messages before pushing.

chaubold:
>That's exactly what it does @chaubold! 

Great! Then I'll play around with this soon :)

>What it doesn't provide though, is the functionality of stacking datasets from different files. Is that something I should look into?

You referenced https://github.com/ilastik/ilastik/issues/1361, where @VolkerH wrote the following, which I understood as separate files, not separate datasets in one HDF5 file (right @VolkerH?):

> we have a microscope that exports data as a sequence of hdf5 files. Each file contains a hdf5 dataset with a 3D-array representing the x,y,z-dimensions.

I think it would be great if we could provide both kinds of importing sequences from HDF5.

k-dominik:
> I think it would be great if we could provide both kinds of importing sequences from HDF5.

Ok, I think it should be possible to add this functionality to the `opStreamingHdf5SequenceReader`.

VolkerH:
> You referenced ilastik/ilastik#1361, where @VolkerH wrote the following, which I understood as separate files, not separate datasets in one HDF5 file (right @VolkerH?):

This is correct. Some microscopes output one hdf5 file (not dataset) per time-point of a time-lapse experiment.

This has a few advantages over having them all in a single hdf5 files as one can already start transferring the individual files to a network share between time points. (Capture often has to be to the HDD/RAID of the acquistion PC due to the high readout speeds of the camera).

Also, the "Big Data Viewer" format is gaining popularity and it is also based on indiviudal hdf5 files. 

Does this newly merged pull request provide this ? Would be great.

k-dominik:
> Does this newly merged pull request provide this ? Would be great.

Hi @VolkerH, I'm still working on it - but yes, this functionality will be implemented.

k-dominik:
Hi, so the pull request is now ready to be reviewed. Both functionalities, reading a sequence of images in a single hdf5 file, as well as reading a sequence of images from multiple hdf5 files, are implemented and tested.

 `OpInputDataReader` selects the appropriate reader by checking the supplied `FilePath`. See the list below for example filepaths. Two new readers were implemented:

* `OpStreamingHdf5SequenceReaderS`: for composing a single volume out of multiple subvolumes inside a single Hdf5-file.
  A suitable file-path contains a string like the following: 
  * suitable for globbing: `/somedir/somefile.h5/some/internal/.path-*`, or
  * list to different subvolumes: `/somedir/somefile.h5/volumes/subvols-00:/somedir/somefile.h5/volumes/subvols-01:...`, 
 
* `OpStreamingHdf5SequenceReaderM`: for composing a single volume out of multiple volumes inside a multiple Hdf5-files.
  A suitable file-path contains a string like the following: 
  * suitable for globbing: `/somedir/somefile-*.h5/some/internal/path`, or
  * list to different subvolumes: `/somedir/somefile-00.h5/volumes/subvol:/somedir/somefile-01.h5/volumes/subvol:...`, 

the respective static classmethod `checkGlobString` is used to determine if the supplied `FilePath` is valid for the readers.

All recent changes from Master have been merged without conflicts.

lazyflow/operators/ioOperators/opStreamingHdf5SequenceReaderM.py
```diff
@@ -0,0 +1,240 @@
...
+        * placeholder '*' in external path
+        * must not contain placeholders in internal paths (for now)
+
+        Args:
+            globString (string): String, one or multiple paths separated with
+              os.path.pathsep and possibly containing '*' as a placeholder.
+
+        Returns:
+            bool: True if rules are met -> this is the right reader
+                  False if rules are not met -> this is not the right reader
```
stuarteberg: Looks like this docstring is incorrect, since this function never returns `False`, AFAICT.

On a related note, this function (and the analogous one in `OpStreamingHdf5SequenceReaderS`) seems a little weird to me.  It mixes modes of "returning" it's result.  It either **returns** `True` or **raises** an exception.  I think it would be clearer if it behaved in one of the following ways:

1. Return nothing, but raise an exception if there's a problem with the `globString`.  (Returning `True` gives the reader the incorrect impression that this function that the return value is meaningful.)

2. Never raise an exception, but instead simply return `True` or `False` depending on whether or not there's a problem.  Of course, the problem with this approach is that the caller would be given no hint as to *why* the globString is bad.

3. Never raise an exception, but instead return some enum-like code to distinguish between `VALID`, `INVALID_WRONG_FILE_TYPE`, `INVALID_NO_EXTERNAL_PLACEHOLDER`, `INVALID_SAME_FILE`.

I suppose option 1 is the simplest fix.

tests/testOpStreamingHdf5SequenceReaderS.py
```diff
@@ -0,0 +1,189 @@
...
+            vigra.AxisInfo("y", typeFlags=vigra.AxisType.Space),
+            vigra.AxisInfo("x", typeFlags=vigra.AxisType.Space),
+            vigra.AxisInfo("c", typeFlags=vigra.AxisType.Channels)])
+
+        expected_axistags = vigra.AxisTags([
+            vigra.AxisInfo("z", typeFlags=vigra.AxisType.Space),
+            vigra.AxisInfo("y", typeFlags=vigra.AxisType.Space),
+            vigra.AxisInfo("x", typeFlags=vigra.AxisType.Space),
+            vigra.AxisInfo("c", typeFlags=vigra.AxisType.Channels)])
+
```
k-dominik: Oh, thanks for the tip, appreciated!

lazyflow/operators/ioOperators/opStreamingHdf5SequenceReaderM.py
k-dominik: I will fix it as proposed and update this PR. Thanks a lot!

k-dominik:
Hi, so I am not sure what the matter is with CircleCi - but locally all tests pass. @stuarteberg: your suggested changes were implemented.

AlexisMaizel:
@VolkerH we faced the same issue and we are happy to read that  reading a sequence of images from multiple hdf5 files is now implemented.
I totally support your point about the "BigDataViewer" H5 format  which we do use. The ability to go from it directly to ilastik would be great.

chaubold:
@VolkerH and @AlexisMaizel: that's a good point about the BigDataViewer format, we'll look into that. In the meantime we've been working on the other way (which will be released as ImageJ plugin very soon): we're adding a BigDataViewer reader for the HDF5 format that ilastik outputs. So stay tuned, we're trying to make your life easier 😉 .

PR #230, opened 2017-03-23
BypassCache: checking for BypassModeEnabled first

JaimeIvanCervantes:
`_execute_Output_impl` was being called wven when `BypassModeEnabled` was set to `True`. I've decide to check `BypassModeEnabled` first before checking for anything else.


lazyflow/operators/opSimpleBlockedArrayCache.py
```diff
@@ -57,19 +57,19 @@ def copy_block( full_block_roi, clipped_block_roi ):
             # If data data exists already or we can just fetch it without needing extra scratch space,
             # just call the base class
```
stuarteberg: The comment above was intended to explain the first branch of this if statement.  Now that you've moved that branch down a few lines, please also move the comment down there (line 69).

JaimeIvanCervantes: Ups. Fixed.

PR #231, opened 2017-03-24
Watershed workflow

shaker6691:
Corrections by stuarteberg and the function: clearAllLabels to clear the LabelCache as fast as possible.


k-dominik:
Okay, to me it looks like we can safely discard this. @akreshuk any reason to keep this in? And just for my understanding, does watershed workflow refer to carving here?

PR #233, opened 2017-04-06
Disabled compression in labelled volume cache

JaimeIvanCervantes:
Disabled the cache compression in `opLabelVolume `:

For a video with 1000 frames, the running-time improves from 365 seconds to 243 seconds.

### DIFF
--- a/lazyflow/operators/opLabelVolume.py
+++ b/lazyflow/operators/opLabelVolume.py
@@ -207,7 +207,6 @@ class OpLabelingABC(Operator):
         self._cache = OpBlockedArrayCache(parent=self)
         self._cache.name = "OpLabelVolume.OutputCache"
         self._cache.BypassModeEnabled.connect(self.BypassModeEnabled)
-        self._cache.CompressionEnabled.setValue(True)
         self._cache.Input.connect(self.Output)
         self.CachedOutput.connect(self._cache.Output)
         self.CleanBlocks.connect(self._cache.CleanBlocks)


stuarteberg:
This was breaking the tests.  Did you figure that out before merging?

JaimeIvanCervantes:
The test was failing with my previous commit: https://github.com/ilastik/lazyflow/commit/d9e5c1b89dc25f963b5f9db5be0a67be7a5abaca

Working on it.

akreshuk:
How does that affect memory consumption?

JaimeIvanCervantes:
I ran a test with 30k frames and the RAM is behaving. Remember that we switched the new `OpBlockedArrayCache` which is regulated by the `cacheMemoryManager`. In the past we were using an old compressed cache.

Also, when running tracking in batch mode, the cache is not even used, so the compression is only affecting the speed when the user is on the GUI.

On another note, the volumes are already compressed when saved to the ilp project file.

akreshuk:
I'm worried because label volume is 4 times the original data size, but on the other hand, if the cache is managed correctly is should not keep the unneeded time slices in memory anyway. I'm not sure how exactly the memory consumption of object classification behaves now after the cache changes. But perhaps it would make sense to enable the compression if we detect that there is not much RAM available and the input data is big? Not for you to do, just thinking aloud.

JaimeIvanCervantes:
The `cacheMemoryManager` should be able to handle the RAM, even if it's 4x the original volume. On the other hand the improvement in object feature computation is about 1.5x, which is significant.

Also, if you run into any trouble in the future, you can just revert this commit ;).

PR #238, opened 2017-05-05
Bug/tifffile wo axes tags

k-dominik:
this fixes issue https://github.com/ilastik/ilastik/issues/1487

basically copied the axis order inference from `OpStreamingHdf5Reader`. Also added a test.

### DIFF
--- a/lazyflow/utility/helpers.py
+++ b/lazyflow/utility/helpers.py
@@ -250,7 +250,45 @@ class newIterator:
         retSlice = [(self.toSlice(src,True),self.toSlice(trgt,True),self.toSlice(mask)) \
                     for src,trgt,mask in retRoi]
         return retSlice.__iter__()
- 
+
+
+def get_default_axisordering(shape):
+    """Given a data shape, return the default axis ordering
+
+    For data types that do not support axistags, we assume a default axis
+    ordering, given the shape, and implicitly the number of dimensions.
+
+    Args:
+        shape (tuple): Shape of the data
+
+    Returns:
+        str: String, each position represents one axis.
+    """
+    axisorders = {
+        2: 'yx',
+        3: 'zyx',
+        4: 'zyxc',
+        5: 'tzyxc'
+    }
+    ndim = len(shape)
+
+    if ndim in [0, 1]:
+        raise ValueError(
+            "Got 'ndim' == {dim}. {dim}-D data not yet supported".format(dim=ndim))
+    elif ndim > 5:
+        raise ValueError(
+            "Got 'ndim' == {dim} dim. No Support for data with more than 5 "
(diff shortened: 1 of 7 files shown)


k-dominik:
Ok, I think I might have forgotten to push my latest commit. I'm afraid it'll have to wait until Monday...

k-dominik:
Okay. I have looked into this again. The test fails on circleCI because dependencies for `ilastik-deps-headless` (on circleCI) and `ilastik-everything-no-solvers` (on my local machines) are not in sync. Particularly the versions of scikit-image differ (and also tifffile) (0.13.0 (remote) and 0.12.3 (local)). In the newer version, on circle-CI, the file is created with valid axes tags ('tyx') and the problem doesn't occur in the first place.
 
In conclusion, before merging this PR the following things should be done

For `testOpTiffReader.py`:
- [ ] Write the test such that it checks for the scikit-image version and always saves a file with unknown axis tags
- [ ] Test all different `ndims`

and also 

- [ ] the code to determine the axis-order from `ndims` can be found multiple times in lazyflow. This should be moved to a common place (`lazyflow.utility`?!)

lazyflow/utility/helpers.py
```diff
@@ -250,7 +250,45 @@ def __iter__(self):
...
+
+    Returns:
+        str: String, each position represents one axis.
+    """
+    axisorders = {
+        2: 'yx',
+        3: 'zyx',
+        4: 'zyxc',
+        5: 'tzyxc'
+    }
```
stuarteberg: I'm not sure, but I think it's more common to see tiff files that use `cyx` order instead of interleaved channels (`yxc`).  Similarly, I think `tzcyx` or `tczyx` might be more common than ilastik's internal default of `tzyxc`.

(I wonder if we can find any example multi-channel files in the wild to corroborate or dispute this claim of mine....)

k-dominik: Hi @stuarteberg,

thank you for looking at the PR.

I have used what I have found in the lazyflow code, so the default behaviour _now_ is unchanged.
At least we can control it from a single point now. Should we leave the PR open until we can have a discussion on that?

Or, would you say, that default axis ordering depends on the file type and should be handled more flexibly?

akreshuk: > At least we can control it from a single point now.
Amen to that. 

We want to use tzyxc internally, but how exactly we interpret the input is open. I just read some random big tiff i found on my hdd with vigra, it turned out as xyzc. Another one with colors was xyc, but maybe it's vigra default order? How would you check the order "in the wild"? Interestingly, if you try to read a 2D RGB image with vigra.readVolume, it just eats RAM and hangs. I wonder what causes that.

PR #249, opened 2017-08-21
Added test for opBigTiffReader

k-dominik:
Found that `OpBigtiffReader` was untested, so I have added a basic test.

Checks if a bigtiff-file, generated with pytiff, can be read.


k-dominik:
thank you!

PR #266, opened 2018-04-23
Circle switch to 2

k-dominik:
Migrated to circleci 2.0, will clean it up before merge, but wanted to see if it properly builds.


k-dominik:
This seems to be working fine, so I hope that resolves our problems here.

PR #269, opened 2018-05-03
raise an exception immediately if localhost:8888 is occupied

FynnBe:
before assertions in testopTiledVolumeReader failed (with logged warnings), due to zero filled output, now the hardcoded port is checked in advance

### DIFF
--- a/tests/testTiledVolume.py
+++ b/tests/testTiledVolume.py
@@ -81,9 +81,13 @@ class DataSetup(object):
         global port
         try:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+                # allow the socket port to be reused if in TIME_WAIT state
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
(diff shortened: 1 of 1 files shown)


FynnBe:
@k-dominik, do you know why this fails on CircleCI? Is there a parallelization issue?

k-dominik:
This port is already in use on circlrci. Can't we just use a different one? 8889 ?

FynnBe:
How did the tests pass in the past then? 8888 is hardcoded!?

k-dominik:
Good point. Let me look into it when I have access to a computer

FynnBe <notifications@github.com> schrieb am Do., 3. Mai 2018, 20:13:

FynnBe:
Indeed that is what happened. Yes 8888 is a bad choice. Dynamically would be best

k-dominik:
hmmm, not sure how overengineered it is for our use-case, but the jupyter guys do the dynamic port thingy like this:

https://github.com/jupyter/notebook/blob/9d4a2da36208ec3d09294c8b41e102fda033f7c9/notebook/notebookapp.py#L1319-L1338

PR #270, opened 2018-05-19
Simplify vigra ops

FynnBe:
will break ilastik tests


FynnBe:
`RuntimeError: VigraArray.withAxes(): array must not contain axes of unknown type (key '?').`
is due to a bug in fastfilters, which is fixed in this PR: https://github.com/ilastik/fastfilters/pull/3 (just needs merging)

lazyflow/operators/ioOperators/ioOperators.py
```diff
@@ -42,6 +42,70 @@
 from lazyflow.request import Request
 
 
+class OpImageReader(Operator):
```
k-dominik: aye, makes way more sense here

FynnBe:
merged with merge commit (not rebase & merge) in https://github.com/ilastik/lazyflow/pull/273

PR #271, opened 2018-05-26
Error using pixel classification with 2D h5 files

hafizfahad143:
issue #1749 solved that open in ilastik but i found the bug in lazyflow

during  2d Data in pixelClassificationDataExportGui.py file show error  that 'c' is not found that why i put a condition if 'c' is not found then add a channel during hdf5 reader and also 'c' in axistags
and in the end  execute function we already change the dimension and axis info of the data. but when we read again data from orignal hdf5 file i face some error thats why i comeback to original shape to execute the data

### DIFF
--- a/lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
+++ b/lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
@@ -86,15 +86,14 @@ class OpStreamingHdf5Reader(Operator):
         assert len(axistags) == len( dataset.shape ),\
             "Mismatch between shape {} and axisorder {}".format( dataset.shape, axisorder )
 
-        if 'c' is not axisorder:
-            self.OutputImage.meta.dtype = dataset.dtype.type
-            self.OutputImage.meta.shape = (dataset.shape[0], dataset.shape[1], 1)
+        self.OutputImage.meta.dtype = dataset.dtype.type
+        if 'c' not in axisorder:
+            self.OutputImage.meta.shape = (*dataset.shape, 1)
             axisorder = get_default_axisordering(self.OutputImage.meta.shape)
-            axistags = vigra.defaultAxistags(str(axisorder))
+            axistags = vigra.defaultAxistags(axisorder)
             self.OutputImage.meta.axistags = axistags
         else:
             # Configure our slot meta-info
-            self.OutputImage.meta.dtype = dataset.dtype.type
             self.OutputImage.meta.shape = dataset.shape
             self.OutputImage.meta.axistags = axistags
 
@@ -130,17 +129,20 @@ class OpStreamingHdf5Reader(Operator):
             timer = Timer()
             timer.unpause()        
 
-        if result.flags.c_contiguous:
-            hdf5File[internalPath].read_direct( result[...], key )
-        else:
-            if len(hdf5File[internalPath].shape) == 2:
-                roi.start = (roi.start[0], roi.start[1])
-                roi.stop = (roi.stop[0], roi.stop[1])
-                key = roi.toSlice()
-                result = result[:, :, 0]
-                result[...] = hdf5File[internalPath][key]
+        shapediff = len(hdf5File[internalPath].shape) - len(result.shape)
+        assert shapediff == 0 or shapediff == -1
+        if shapediff == -1:
+            roi.start = roi.start[:-1]
+            roi.stop = roi.stop[:-1]
+            key = roi.toSlice()
+            tmp_result = hdf5File[internalPath][key]
+            result[...] = tmp_result[..., numpy.newaxis]
+        elif shapediff == 0:
+            if result.flags.c_contiguous:
+                hdf5File[internalPath].read_direct(result[...], key)
             else:
                 result[...] = hdf5File[internalPath][key]
+
         if logger.getEffectiveLevel() >= logging.DEBUG:
             t = 1000.0*(time.time()-t)
             logger.debug("took %f msec." % t)
--- a/lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
+++ b/lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
@@ -82,21 +82,20 @@ class OpStreamingHdf5Reader(Operator):
             # No axistags found.
             axisorder = get_default_axisordering(dataset.shape)
             axistags = vigra.defaultAxistags(str(axisorder))
+            axisorder += 'c'
 
         assert len(axistags) == len( dataset.shape ),\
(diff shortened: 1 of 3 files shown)


lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
```diff
@@ -86,10 +86,17 @@ def setupOutputs(self):
         assert len(axistags) == len( dataset.shape ),\
             "Mismatch between shape {} and axisorder {}".format( dataset.shape, axisorder )
 
-        # Configure our slot meta-info
-        self.OutputImage.meta.dtype = dataset.dtype.type
-        self.OutputImage.meta.shape = dataset.shape
-        self.OutputImage.meta.axistags = axistags
+        if 'c' is not axisorder:
+            self.OutputImage.meta.dtype = dataset.dtype.type
+            self.OutputImage.meta.shape = (dataset.shape[0], dataset.shape[1], 1)
```
k-dominik: Your code will only work with 2D data. To make it work with more dims, please use:

`(*dataset.shape, 1)`, to append a channel axis

lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
```diff
@@ -126,7 +133,14 @@ def execute(self, slot, subindex, roi, result):
         if result.flags.c_contiguous:
             hdf5File[internalPath].read_direct( result[...], key )
         else:
-            result[...] = hdf5File[internalPath][key]
+            if len(hdf5File[internalPath].shape) == 2:
```
k-dominik: here, also to make it work with nd, it should probably be something like: 

```
shapediff = len(hdf5File[internalPath].shape) - result.shape
assert shapediff == 0 or shapediff == -1
if shapediff == -1:
    # code for adding the channel
elif shapediff == 0:
    # shapes match, here goes the code from the original implementation

File: lazyflow/operators/ioOperators/opStreamingHdf5Reader.py [hunk 3]
```diff
@@ -126,7 +133,14 @@ def execute(self, slot, subindex, roi, result):
         if result.flags.c_contiguous:
```
k-dominik: this check should be done if the shapes match only

lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
```diff
@@ -126,7 +133,14 @@ def execute(self, slot, subindex, roi, result):
...
             hdf5File[internalPath].read_direct( result[...], key )
         else:
-            result[...] = hdf5File[internalPath][key]
+            if len(hdf5File[internalPath].shape) == 2:
+                roi.start = (roi.start[0], roi.start[1])
```
k-dominik: Here we now assume, that it can have more than 2d, so 
`roi.start = roi.start[:-1]  # this excludes the last element, what we expect to be the channel, same for roi.stop

lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
```diff
@@ -126,7 +133,14 @@ def execute(self, slot, subindex, roi, result):
...
+                roi.stop = (roi.stop[0], roi.stop[1])
+                key = roi.toSlice()
+                result = result[:, :, 0]
```
k-dominik: I wonder that this works, doesn't that change the shape of result? In any case you should rather do something like the following, and really return the result in the right shape (with an added channel axis)

```
tmp_result = hdf5File[internalPath][key]
result[...] = tmp_result[..., numpy.newaxis]
```

hafizfahad143: yes it works but i am changing with this one. thanks

lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
hafizfahad143: i didn't understand this point

k-dominik:
hmm, yes, could you please rebase this branch onto master and force push it again?


```
# on lazyflow master branch
# assuming upstream is configured to be ilastik/lazyflow
git fetch upstream master
git rebase upstream/master

# then switch to your feature branch, rebase it onto the current master and force push the result
git checkout pixel-clasification-2d-image-error
git rebase master
git push --force
```

just in order to include recent fixes, thanks!

hafizfahad143:
done it .

k-dominik:
There seem to be some legitimate test failures in `testOpFormattedDataExport.TestOpFormattedDataExport.testBasic` and . have you run the tests yourself?

you can do that by

```
# assuming you are in ilastik-meta/lazyflow
nosetests --where=./tests
```
you should see then, that two of the tests fail.

I think this has something to do with the additional channel, that is added to data without a channel.

Please also note that due some recent changes you have to update the `fastfilters` package in your conda environment:

```
# assuming you have activated the environment
conda update -c ilastik-forge fastfilters 
```

lazyflow/operators/ioOperators/opStreamingHdf5Reader.py
```diff
@@ -86,10 +86,16 @@ def setupOutputs(self):
...
         self.OutputImage.meta.dtype = dataset.dtype.type
-        self.OutputImage.meta.shape = dataset.shape
-        self.OutputImage.meta.axistags = axistags
+        if 'c' not in axisorder:
+            self.OutputImage.meta.shape = (*dataset.shape, 1)
+            axisorder = get_default_axisordering(self.OutputImage.meta.shape)
+            axistags = vigra.defaultAxistags(axisorder)
+            self.OutputImage.meta.axistags = axistags
+        else:
+            # Configure our slot meta-info
```
k-dominik: this comment should come before the `if 'c' not in axisorder:` line, as in both branches, the meta-info is configured

tests/testOpExportSlot.py
```diff
@@ -79,8 +79,8 @@ def testBasic_Hdf5(self):
             opRead.cleanUp()
 
     def testBasic_Npy(self):
-        data = numpy.random.random( (100,100) ).astype( numpy.float32 )
-        data = vigra.taggedView( data, vigra.defaultAxistags('xy') )
+        data = numpy.random.random( (100,100,1) ).astype( numpy.float32 )
```
k-dominik: please add a space after the comma, please do so also for the other changed line with the `CoordinateOffset`

k-dominik:
fixes https://github.com/ilastik/ilastik/issues/1845, https://github.com/ilastik/ilastik/issues/1749

k-dominik:
on hold for now, as this breaks some things in ilastik, mostly tests

PR #272, opened 2018-05-26
make sure original axistags always have a channel axis

FynnBe:
When exporting data, the original axistags are used. If the input data has no explicit channel axis, but the data to be exported has, an error is thrown. Always adding a channel axis prevents this.

Breaks ilastik tests, fixed by: https://github.com/ilastik/ilastik/pull/1774


akreshuk:
You fixed it before I submitted it! Found this annoying behaviour earlier this week. The fix looks good to me.

PR #279, opened 2018-09-17
Import/Export N5 files

paulhfu:
This fixes ilastik/ilastik#1788


lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
```diff
@@ -50,30 +54,59 @@ class OpStreamingHdf5Reader(Operator):
     OutputImage = OutputSlot()
 
     H5EXTS = ['.h5', '.hdf5', '.ilp']
+    N5EXTS = ['.n5']
+
+    # monkeypatch z5py to have it similar enough to be used like h5py
```
paulhfu: @k-dominik, do you think this is a good place to do the monkeypatching for z5py? I could also try to move it all to __init__

paulhfu:
Hey @k-dominik, this and the corresponding, related PR's in ilastik and volumina are ready to be reviewed.

lazyflow/operators/ioOperators/ioOperators.py
```diff
@@ -559,51 +561,54 @@ def setupOutputs(self):
...
-        self.chunkShape = determineBlockShape( list(tagged_maxshape.values()), 512000.0 / dtypeBytes )
+        self.chunkShape = determineBlockShape(list(tagged_maxshape.values()), 512000.0 / dtypeBytes)
 
         if datasetName in list(g.keys()):
             del g[datasetName]
-        kwargs = { 'shape' : dataShape, 'dtype' : dtype,
-            'chunks' : self.chunkShape }
+        kwargs = {'shape': dataShape, 'dtype': dtype,
+                  'chunks': self.chunkShape}
         if self.CompressionEnabled.value:
```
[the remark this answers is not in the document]
k-dominik: So things at z5 are moving fast. Turns out they do have compression enabled per default because they target really large data. In any case, with the two menu entries we have, the uncompressed option should really be uncompressed :)
so this `if`, needs and `else`, or just define some default `kwargs` without compression before the if clause

lazyflow/operators/ioOperators/opExportSlot.py
```diff
@@ -236,40 +240,46 @@ def run_export(self):
...
         export_components = PathComponents(self.ExportPath.value)
         try:
-            os.remove(export_components.externalPath)
+            if export_components.extension in OpStreamingH5N5Reader.N5EXTS:
+                if os.path.isdir(export_components.externalPath):
+                    shutil.rmtree(export_components.externalPath)
+                else:
+                    os.remove(export_components.externalPath)
+            else:
+                os.remove(export_components.externalPath)
```
k-dominik: I'm a bit confused by this - sorry. So, we have two possibilities here, right?

a) `n5` file: -> delete a folder
b) `hdf5` file: -> delete a file

but in a) you also seem to be expecting files as external paths? Or is it for the case that, for whatever reason someone actually had a single file there with the `.h5` extension?

lazyflow/operators/ioOperators/opExportSlot.py
```diff
@@ -236,40 +240,46 @@ def run_export(self):
...
+
+    def _export_h5n5(self, compress=False):
         self.progressSignal( 0 )
 
-        # Create and open the hdf5 file
+        # Create and open the hdf5/n5 file
         export_components = PathComponents(self.ExportPath.value)
         try:
-            os.remove(export_components.externalPath)
+            if export_components.extension in OpStreamingH5N5Reader.N5EXTS:
```
k-dominik: Here I see the potential danger, that `OpStreamingH5N5Reader.N5EXT` and `self._export_impls` might get out of sync...

lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
k-dominik: monkey patching should happen as early as possible. how about `lazyflow/__init__.py` ? Just to make sure all our submodules see the same `z5py`.
By the way, have you checked back with constantin, whether the monkey-patching is still necessary?
I've build a new version of `z5py` which is available via `ilastik-forge`...

lazyflow/utility/pathHelpers.py
```diff
@@ -327,11 +329,11 @@ def mkdir_p(path):
             raise
 
 
-def lsHdf5(hdf5FileObject, minShape=2, maxShape=5):
-    """Generates dataset list of given h5py file object
+def lsH5N5(h5N5FileObject, minShape=2):
```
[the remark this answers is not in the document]
paulhfu: Ah, sorry I wanted to ask you that and forgot it. I will undo it. I removed it here because the other function (which i removed) _get_hdf5_dataset_names() in opInputDataReader.py also did not have it...

lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
paulhfu: I checked in z5py and it is still necessary. If I recall it correctly, Constantin and I decided to leave it like that. I moved it to ```lazyflow/__init__.py```

constantinpape: I think the issues were `read_direct` and `write_direct`, which, as far as I understand, reimplement `[]` with limited slicing support in `h5py`, while potentially avoiding a memory copy.
I don't support most of the more fancy slicing in `z5py`, so this functionality is basically the same as `[]`.
That being said, we could still implement this in `z5py`. You could just make a PR with the functions below @paulhfu.

tests/testOpStreamingH5N5SequenceReaderM.py
```diff
@@ -0,0 +1,250 @@
+import unittest
```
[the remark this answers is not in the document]
paulhfu: check

lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
```diff
@@ -50,30 +54,59 @@ class OpStreamingHdf5Reader(Operator):
...
+    def z5pyReadDirectWrapper(self, dest, source_sel=None, dest_sel=None):
+        start = [s.start for s in source_sel]
+        stop = [s.stop for s in source_sel]
+        dest[...] = self.read_subarray(start, stop)
+
+    def z5pyWriteDirectWrapper(self, source, source_sel=None, dest_sel=None):
+        start = [s.start for s in dest_sel]
+        self.write_subarray(start, source)
+
+    z5py.dataset.Dataset.compression = None
```
constantinpape: The `compression` attribute is missing in the `z5py.Dataset` API. I will fix that.
However, I don't quite get why you set it to `None` here, is that just a hack to have something?

lazyflow/operators/ioOperators/opExportSlot.py
paulhfu: Yep, thats sloppy. I removed the else-block

lazyflow/operators/ioOperators/opStreamingH5N5Reader.py
paulhfu: if the h5 dataset is comressed, ilastik reads with a multi-process access object. Setting it to None avoids that. I will change this in ilastik so this only applies to h5

constantinpape:
I just released z5py 1.3.0. With this the monkey patches should be obsolete.

tests/testOpH5N5WriterBigDataset.py
```diff
@@ -0,0 +1,251 @@
...
+import vigra
+import h5py
+import z5py
+import os
+import sys
+import lazyflow.graph
+
+import logging
+#logger = logging.getLogger(__name__)
+#logger.addHandler(logging.StreamHandler(sys.stdout))
```
k-dominik: these commented out portions can most  likely be removed.

tests/testOpH5N5WriterBigDataset.py
```diff
@@ -0,0 +1,251 @@
...
+        h5_dataset = hdf5File[self.datasetInternalPath]
+        n5_dataset = n5File[self.datasetInternalPath]
+        assert h5_dataset.shape == self.dataShape
+        assert n5_dataset.shape == self.dataShape
+        assert numpy.all(h5_dataset[...] == self.testData.view(numpy.ndarray)[...])
+        assert numpy.all(n5_dataset[...] == self.testData.view(numpy.ndarray)[...])
+        hdf5File.close()
+        n5File.close()
+
+if __name__ == "__main__":
```
k-dominik: please remove the `__main__` portion, we don't have those anymore (or will not, soon)

k-dominik: this goes, of course, too, for the other test-files you touched :)

tests/testOpStreamingH5N5SequenceReaderS.py
```diff
@@ -0,0 +1,286 @@
...
+import z5py
+import numpy
+
+from lazyflow.graph import Graph
+from lazyflow.operators.ioOperators import OpStreamingH5N5SequenceReaderS
+import vigra
+
+
+@contextlib.contextmanager
+def tempdir():
```
emilmelnikov: Maybe use `tempfile.TemporaryDirectory` from stdlib?

lazyflow/operators/ioOperators/ioOperators.py
```diff
@@ -526,31 +528,32 @@ def execute(self, slot, subindex, roi, result):
...
     WriteImage = OutputSlot()
 
-    loggingName = __name__ + ".OpH5WriterBigDataset"
+    loggingName = __name__ + ".OpH5N5WriterBigDataset"
     logger = logging.getLogger(loggingName)
     traceLogger = logging.getLogger("TRACE." + loggingName)
 
     def __init__(self, *args, **kwargs):
-        super(OpH5WriterBigDataset, self).__init__(*args, **kwargs)
+        super(OpH5N5WriterBigDataset, self).__init__(*args, **kwargs)
```
emilmelnikov: Now we can just `super().__init__(...)` in Python 3.

lazyflow/operators/ioOperators/opInputDataReader.py
```diff
@@ -340,11 +339,11 @@ def _attemptOpenAsStack(self, filePath):
...
             return ([], None)
 
-    def _attemptOpenAsHdf5(self, filePath):
-        # Check for an hdf5 extension
+    def _attemptOpenAsH5N5(self, filePath):
+        # Check for an hdf5 or n5 extension
         pathComponents = PathComponents(filePath)
         ext = pathComponents.extension
-        if ext not in (".%s" % x for x in OpInputDataReader.h5Exts):
+        if ext not in (".%s" % x for x in OpInputDataReader.h5_n5_Exts):
```
emilmelnikov:
```suggestion
        if ext[1:] not in OpInputDataReader.h5_n5_Exts:
```

lazyflow/__init__.py
```diff
@@ -32,3 +35,19 @@
...
+    """
+    json encoder for json dumps in z5py
+    """
+
+    def default(self, obj):
+        if isinstance(obj, np.integer):
+            return int(obj)
+        if isinstance(obj, np.floating):
+            return float(obj)
+        json.JSONEncoder.default(self, obj)
```
emilmelnikov:
```suggestion
        return json.JSONEncoder.default(self, obj)
```

lazyflow/operators/ioOperators/ioOperators.py
```diff
@@ -559,51 +562,58 @@ def setupOutputs(self):
...
-            if hdf5GroupName in self.f:
-                g = self.f[hdf5GroupName]
+            if h5N5GroupName in self.f:
+                g = self.f[h5N5GroupName]
             else:
-                g = self.f.create_group(hdf5GroupName)
+                g = self.f.create_group(h5N5GroupName)
 
         dataShape=self.Image.meta.shape
         self.logger.info( "Data shape: {}".format(dataShape))
```
emilmelnikov:
```suggestion
         self.logger.info( f"Data shape: {dataShape}")
```

Ditto for other string formats.

lazyflow/operators/ioOperators/opStreamingH5N5SequenceReaderM.py
```diff
@@ -155,22 +168,32 @@ def setupOutputs(self):
...
+                        raise OpStreamingH5N5SequenceReaderM.InconsistentDType(external_path, internal_path)
+                    if shape != h5N5File[internal_path].shape:
+                        raise OpStreamingH5N5SequenceReaderM.InconsistentShape(external_path, internal_path)
                 opReader.InternalPath.setValue(internal_path)
-                opReader.Hdf5File.setValue(hdfFile)
+                opReader.H5N5File.setValue(h5N5File)
             except RuntimeError as e:
                 logger.error(str(e))
-                raise OpStreamingHdf5SequenceReaderM.FileOpenError(external_path)
+                raise OpStreamingH5N5SequenceReaderM.FileOpenError(external_path)
```
emilmelnikov:
```suggestion
                raise OpStreamingH5N5SequenceReaderM.FileOpenError(external_path) from e
```

lazyflow/utility/pathHelpers.py
```diff
@@ -340,37 +342,43 @@ def lsHdf5(hdf5FileObject, minShape=2, maxShape=5):
     listOfDatasets = []
 
     def addObjectNames(objectName, obj):
-        if isinstance(obj, h5py.Dataset):
-            if (len(obj.shape) >= minShape) and (len(obj.shape) <= maxShape):
+        if isinstance(obj, h5py._hl.dataset.Dataset) or isinstance(obj, z5py.dataset.Dataset):
```
emilmelnikov: Can be somewhat simplified by composite `isinstance`check and early returns.
```python
if not isinstance(obj, (h5py._hl.dataset.Dataset, z5py.dataset.Dataset)):
    return
if len(obj.shape) not in range(minShape, maxShape + 1):
    return
```

tests/testOpStreamingH5N5Reader.py
```diff
@@ -21,65 +21,79 @@
...
         self.data = numpy.indices(datashape).sum(0).astype(numpy.float32)
 
     def tearDown(self):
         self.h5File.close()
-        try:
-            os.remove(self.testDataFileName)
-        except:
-            pass
+        self.n5File.close()
+        shutil.rmtree(self.testFileDir)
```
emilmelnikov:
```suggestion
        self.testFileDir.cleanup()
```

lazyflow/__init__.py
k-dominik: nice catch

paulhfu:
Hey @emilmelnikov, thanks for the review. Concerning the separation of h5 and n5. The first attempt actually was that we had separate classes for each but then decided to merge them. Of course I can still roll that back.

lazyflow/__init__.py
emilmelnikov: Even better:
```suggestion
        return super().default(obj)
```
Although it doesn't really make a difference (this raises an exception).

emilmelnikov:
> Hey @emilmelnikov, thanks for the review. Concerning the separation of h5 and n5. The first attempt actually was that we had separate classes for each but then decided to merge them. Of course I can still roll that back.

I see, maybe it's better to leave it as-is for now, because it looks like `OpExportSlot` needs to be refactored into separate subclasses and a factory.

emilmelnikov:
It's usually better to use contextmanagers instead of try/finally if the actions in finally are simple, but in this case yes, I don't think it's super necessary to do it.

tests/testOpStreamingH5N5SequenceReaderS.py
```diff
@@ -17,6 +16,9 @@ class TestOpStreamingH5N5SequenceReaderS(unittest.TestCase):
 
     def setup_method(self, method):
         self.graph = Graph()
+        self.tempdir = tempfile.TemporaryDirectory()
```
emilmelnikov: We could use pytest's [`tmp_path` fixture](https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture) and get a temp dir for free. Although I'm not sure how this interacts with the class-based unittest tests...

PR #281, opened 2018-10-30
Switch to travis

k-dominik:
With travis we can test on both, linux and macos for free these days.

Possible alternative would be azure devops services where one can test on all three major platforms


k-dominik:
what do you guys think? is anyone rooting for microsoft? otherwise i'd add travis/remove circleci from the other repos as well

emilmelnikov:
What about GitLab CI? They are very good at DevOps, and their runner tool is cross-platform, open-sourced and can be used in a self-hosted setting if needed. However, I am not sure if they have Windows and macOS instances in their SaaS.

k-dominik:
@emilmelnikov is there a free version of the gitlab CI?

emilmelnikov:
Yes, sure, they are free for open-source projects: https://about.gitlab.com/solutions/open-source/.

PR #290, opened 2019-02-20
Enabling pixel classification mask

Tomaz-Vieira:
# Refining the code for the Op*ClassificationPredict operators.

The Op*ClassificationPredict operators are able to handle masks to determine whether computation was needed on a roi during Pixel Classification. This PR re-applies the mask after the computation is done, so that the user will only get positive predictions  in the areas of the image that are not masked out.

Also, I've removed some of the code replication by moving common logic present in both OpPixelwiseClassifierPredict and OpVectorwiseClassifierPredict into a generic base class.

TODO: It's possible that we could optimize the code some more since we're still calculating features and predicitons for all pixels even if the roi is mostly made out of masked pixels.

### DIFF
--- a/lazyflow/operators/classifierOperators.py
+++ b/lazyflow/operators/classifierOperators.py
@@ -432,9 +432,16 @@ class OpBaseClassifierPredict(Operator):
         mask = None
         if self.PredictionMask.ready():
             mask_roi = numpy.array((roi.start, roi.stop))
-            mask_roi[:,-1:] = [[0],[1]]
+            num_channels_in_mask = self.PredictionMask.meta.getTaggedShape()['c']
+            mask_roi[:,-1:] = [[0],[num_channels_in_mask]]
             start, stop = list(map(tuple, mask_roi))
-            mask = self.PredictionMask( start, stop ).wait()
+            multichannel_mask = self.PredictionMask( start, stop ).wait()
+
+            #create a single-channel merged mask, which has 0 iff all PredictionMask channels are 0
+            mask = multichannel_mask[...,0:1]
+            for c in range(1, num_channels_in_mask):
+                mask = numpy.logical_or(mask, multichannel_mask[..., c:c+1])
+
             if not numpy.any(mask):
                 logger.debug(f"Skipping masked block {roi}")
                 result[:] = 0.0
@@ -457,8 +464,7 @@ class OpBaseClassifierPredict(Operator):
 
         # Cancel out masked pixels.
         if mask is not None:
-            mask_all_channels = numpy.broadcast_to(mask == 0, probabilities.shape)
-            probabilities[mask_all_channels] = 0.0
+            probabilities *= mask
 
         # Copy only the prediction channels the client requested.
         result[...] = probabilities[..., roi.start[-1]:roi.stop[-1]]
@@ -476,7 +482,7 @@ class OpBaseClassifierPredict(Operator):
         elif slot == self.Image:
             self.PMaps.setDirty()
         elif slot == self.PredictionMask:
-            self.PMaps.setDirty(roi.start, roi.stop)
+            self.PMaps.setDirty()
 
 class OpPixelwiseClassifierPredict(OpBaseClassifierPredict):
     def _calculate_probabilities(self, roi):
@@ -554,7 +560,6 @@ class OpVectorwiseClassifierPredict(OpBaseClassifierPredict):
         prod = numpy.prod(shape[:-1])
         features = input_data.reshape((prod, shape[-1]))
 
-        classifier = self.Classifier.value
         with Timer() as prediction_timer:
             probabilities = classifier.predict_probabilities( features )
 
--- a/lazyflow/operators/classifierOperators.py
+++ b/lazyflow/operators/classifierOperators.py
@@ -21,7 +21,7 @@ from __future__ import absolute_import
 #		   http://ilastik.org/license/
 ###############################################################################
 #Python
-from abc import abstractmethod
+from abc import abstractmethod, ABC
 import copy
 import logging
 traceLogger = logging.getLogger("TRACE." + __name__)
@@ -390,7 +390,7 @@ class OpClassifierPredict(Operator):
         if slot == self.Classifier:
             self.PMaps.setDirty()
 
-class OpBaseClassifierPredict(Operator):
+class OpBaseClassifierPredict(Operator, ABC):
     Image = InputSlot()
     LabelsCount = InputSlot()
     Classifier = InputSlot()
@@ -482,7 +482,7 @@ class OpPixelwiseClassifierPredict(OpBaseClassifierPredict):
     def _calculate_probabilities(self, roi):
         classifier = self.Classifier.value
 
-        assert issubclass(type(classifier), LazyflowPixelwiseClassifierABC), \
+        assert isinstance(classifier, LazyflowPixelwiseClassifierABC), \
             f"Classifier {classifier} must be sublcass of {LazyflowPixelwiseClassifierABC}"
 
         upstream_roi = (roi.start, roi.stop)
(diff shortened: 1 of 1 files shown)


lazyflow/operators/classifierOperators.py
```diff
@@ -437,40 +438,15 @@ def execute(self, slot, subindex, roi, result):
             skip_prediction = not numpy.any(mask)
 
         if skip_prediction:
+            logger.debug(f"Skipping masked block {roi}")
```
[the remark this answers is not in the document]
Tomaz-Vieira: Nice catch =)
Done

lazyflow/operators/classifierOperators.py
```diff
@@ -437,40 +438,15 @@ def execute(self, slot, subindex, roi, result):
...
-        # Determine how to extract the data from the result (without the halo)
-        downstream_roi = numpy.array((roi.start, roi.stop))
-        predictions_roi = downstream_roi[:,:-1] - upstream_roi[0,:-1]
+        assert issubclass(type(classifier), self.classifierInterface), \
+            f"Classifier is of type {type(classifier)}, which does not satisfy the {self.classifierInterface} interface."
 
-        # Request all upstream channels
-        input_channels = self.Image.meta.shape[-1]
-        upstream_roi[:,-1] = [0, input_channels]
+        probabilities = self._calculate_probabilities(roi)
```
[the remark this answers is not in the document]
Tomaz-Vieira: I made the mehod abstract, but since the class inherits from operator, which already has a metaclass, I'm not sure I can safely make the class itself abstract =/

lazyflow/operators/classifierOperators.py
```diff
@@ -479,12 +479,12 @@ def propagateDirty(self, slot, subindex, roi):
...
 
 class OpPixelwiseClassifierPredict(OpBaseClassifierPredict):
-    @property
-    def classifierInterface(self):
-        return LazyflowPixelwiseClassifierABC
-
     def _calculate_probabilities(self, roi):
         classifier = self.Classifier.value
+
+        assert issubclass(type(classifier), LazyflowPixelwiseClassifierABC), \
```
emilmelnikov: Would `isinstance(classifier, LazyflowPixelwiseClassifierABC)` work?

lazyflow/operators/classifierOperators.py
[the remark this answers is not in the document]
emilmelnikov: You also can inherit from `abc.ABC` to get an abstract class. Without `abc` magic, `@abstractmethod` will do nothing in runtime, so in that case it's better to at least `raise NotImplementedError`.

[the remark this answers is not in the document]
emilmelnikov: You also can inherit from `abc.ABC` to get an abstract class. Without `abc` magic, `@abstractmethod` will do nothing in runtime, so in that case it's better to at least `raise NotImplementedError`.

lazyflow/operators/classifierOperators.py
Tomaz-Vieira: absolutely. done =)

lazyflow/operators/classifierOperators.py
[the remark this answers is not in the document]
Tomaz-Vieira: Added inheritance do ABC. Maybe in the future we could move that into `Operator` itself

lazyflow/operators/classifierOperators.py
```diff
@@ -437,40 +438,15 @@ def execute(self, slot, subindex, roi, result):
...
-        upstream_roi = numpy.array( upstream_roi )
-        upstream_roi[0] -= halo_shape
-        upstream_roi[1] += halo_shape
-        upstream_roi = getIntersection( upstream_roi, roiFromShape(self.Image.meta.shape) )
-        upstream_roi = numpy.asarray( upstream_roi )
-
-        # Determine how to extract the data from the result (without the halo)
-        downstream_roi = numpy.array((roi.start, roi.stop))
-        predictions_roi = downstream_roi[:,:-1] - upstream_roi[0,:-1]
+        assert issubclass(type(classifier), self.classifierInterface), \
```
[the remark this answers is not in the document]
Tomaz-Vieira: Done.

lazyflow/operators/classifierOperators.py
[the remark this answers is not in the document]
emilmelnikov: Actually, `OperatorMetaClass` already inherits from `ABCMeta`, so you should be able to use `@abstractmethod` here without inheriting from `ABC`. Sorry for the confustion. Although it's strange that `Operator.execute` and friends are not defined as abstract methods.

lazyflow/operators/classifierOperators.py
```diff
@@ -577,48 +550,16 @@ def execute(self, slot, subindex, roi, result):
             input_data = self.Image[newKey].wait()
 
         input_data = numpy.asarray(input_data, numpy.float32)
-
         shape=input_data.shape
         prod = numpy.prod(shape[:-1])
         features = input_data.reshape((prod, shape[-1]))
 
+        classifier = self.Classifier.value
```
k-dominik: `classifier` already exists

```suggestion
```

lazyflow/operators/classifierOperators.py
```diff
@@ -497,36 +478,45 @@ def propagateDirty(self, slot, subindex, roi):
         elif slot == self.PredictionMask:
             self.PMaps.setDirty(roi.start, roi.stop)
```
k-dominik: for simplicity I'd opt to set everything dirty here, too

```suggestion
            self.PMaps.setDirty()
```

lazyflow/operators/classifierOperators.py
```diff
@@ -484,10 +455,20 @@ def execute(self, slot, subindex, roi, result):
             
             probabilities = full_probabilities
 
+        # Cancel out masked pixels.
+        if mask is not None:
+            mask_all_channels = numpy.broadcast_to(mask == 0, probabilities.shape)
```
k-dominik: just a matter of taste, but I'd prefer 
```python
            probabilities *= (mask != 0)
```

instead of 

```python
            mask_all_channels = numpy.broadcast_to(mask == 0, probabilities.shape)
            probabilities[mask_all_channels] = 0.0
```
just because from the documentation ("Broadcast an array to a new shape.") it is not really clear to me what `broadcast_to` should to, how it fails, etc.

Tomaz-Vieira: I like your style better; Far more concise. It's probably doing almost the same behind the scenes, anyway. Since `mask` can have a shape that is different from that of `probabilities`, it still will have to be implicitly broadcast to match `probabilities`'s shape.

Your comment also made me realize that the previous code was misinterpreting the mask, making it work channel-wise. I've changed it to make the mask be  logical `OR` of all the channels, so that a user can have colored masks that will only cause the black pixels to be masked out

lazyflow/operators/classifierOperators.py
Tomaz-Vieira: done

lazyflow/operators/classifierOperators.py
Tomaz-Vieira: nice catch =)

lazyflow/operators/classifierOperators.py
```diff
@@ -417,60 +415,40 @@ def setupOutputs(self):
...
-        assert len(halo_shape) == len( upstream_roi[0] )
-        assert halo_shape[-1] == 0, "Didn't expect a non-zero halo for channel dimension."
-
-        # Expand block by halo, then clip to image bounds
-        upstream_roi = numpy.array( upstream_roi )
-        upstream_roi[0] -= halo_shape
-        upstream_roi[1] += halo_shape
-        upstream_roi = getIntersection( upstream_roi, roiFromShape(self.Image.meta.shape) )
-        upstream_roi = numpy.asarray( upstream_roi )
+            #create a single-channel merged mask, which has 0 iff all PredictionMask channels are 0
```
k-dominik: I would have just required users to use single channel images, but this is so much better (just imaging using multiple, stacked atlasses at the same time...)

PR #291, opened 2019-02-22
WIP - Trying to remove some of the clutter from Lazyflow

Tomaz-Vieira:
Mostly moving code around and removing weird, probably deprecated stuff. I have no idea if this is stable yet.

I'm also trying to review the locking logic in lazyflow, because it's crashing the GUI since slots might try to service requests when unready because of some disconnect event somewhere.

Do let me know if you see something that was nuked but shouldn't, or something that wasn't and should.

Also, if anyone sees this and knows the answer: is there any good reason to have sub-slots have their parent slot (i.e. the slot just above them, with level == subslot.level + 1) acting like its operator?

### DIFF
--- a/lazyflow/slot.py
+++ b/lazyflow/slot.py
@@ -1,11 +1,5 @@
-from builtins import next
-
-from builtins import range
-from builtins import object
 from future.utils import raise_with_traceback
 import sys
-if sys.version_info.major >= 3:
-    unicode = str
 
 ###############################################################################
 #   lazyflow: data flow based lazy parallel computation framework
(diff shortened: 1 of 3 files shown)


lazyflow/slot.py
```diff
@@ -483,143 +490,71 @@ def _handleUpstreamUnready(self, slot):
...
-                if self._type == upstream_slot._type == "output":
-                    upstream_slot.notifyValueChanged(self._sig_value_changed)
+        if self.upstream_slot is upstream_slot and upstream_slot.level == self.level:
+            return
 
-        except:
-            try:
-                raise
-            finally:
-                try:
```
emilmelnikov: Have you figured out what was going on here? I was scared of touching that part...

k-dominik:
seems like this has been stale for a while and never made it out of "wip": @Tomaz-Vieira, do you want to get back to it, or scratch it?

Tomaz-Vieira:
> seems like this has been stale for a while and never made it out of "wip": @Tomaz-Vieira, do you want to get back to it, or scratch it?
Nah, we can scratch it. Some of it has already been tackled by maxim anyway =)

Tomaz-Vieira:
Scratched.

PR #292, opened 2019-02-25
Parallelize multi-page TIFF export

emilmelnikov:
Previously, OpExportMultipageTiff.run_export() requested pages
sequentially. Although each submitted request eventually triggered
parallel computations, at the end of each request threads were idle
because of the sequential step in the run_export().

This commit uses RoiRequestBatch to request multiple pages in parallel.
Because pages have to be written in order, the current implementation
maintains a page buffer that saves out-of-order pages in memory until it
can release a contiguous sequence of pages. This approach can
potentially blow up memory, but very unlikely to happen because the
workload is uniform across requests, and pages are requested in-order.

The other alternative is to write pages as they arrive, and later
specify their order in the OME-XML ImageDescription field. However, this
requires patching the TIFF file field after the data has been written.,
which is more error-prone to implement and maintain.

Exporting probability map as normalized uint8 image from a 572M file [1]
containing 1600x1200 pixels, 3 channels and 104 time points gave
1.28x speedup from 739 to 578 seconds for run_export().

Measuring page request computation time and comparing it to the page
I/O time showed that the page I/O constitutes 0.7% of the total page
processing time: 3.063 mean seconds for page request vs. 0.022
mean seconds for the page I/O.

System: Dell XPS 15 9570
CPU: Intel Core i7-8750H (6 cores, 12 threads)
RAM: 16GB
OS: Ubuntu 18.04
Disk: NVMe

[1]: http://data.ilastik.org/pixel-classification/3d/MC0509C3X100-2.tif

### DIFF
--- a/lazyflow/operators/ioOperators/opExportMultipageTiff.py
+++ b/lazyflow/operators/ioOperators/opExportMultipageTiff.py
@@ -17,246 +17,265 @@
 # See the files LICENSE.lgpl2 and LICENSE.lgpl3 for full text of the
 # GNU Lesser General Public License version 2.1 and 3 respectively.
 # This information is also available on the ilastik web site at:
-#		   http://ilastik.org/license/
+# 		   http://ilastik.org/license/
 ###############################################################################
-from __future__ import division
-from __future__ import print_function
-#from future import standard_library
-#standard_library.install_aliases()
-from builtins import zip
-from builtins import next
-#
-from builtins import range
-import os
-import collections
-
-import numpy
-import psutil
-
-import vigra
-import tifffile
-
-from lazyflow.graph import Operator, InputSlot
-from lazyflow.utility import OrderedSignal
-from lazyflow.operators.opReorderAxes import OpReorderAxes
 
 import logging
+import pathlib
+import textwrap
+import uuid
+import xml.etree.ElementTree as ET
+from typing import Any, Iterable, Sequence
+
+import numpy as np
+import psutil
+import tifffile
+import vigra
+
+from lazyflow.graph import InputSlot, Operator
+from lazyflow.operators.opReorderAxes import OpReorderAxes
+from lazyflow.utility import OrderedSignal, RoiRequestBatch
+
 logger = logging.getLogger(__name__)
 
+
 class OpExportMultipageTiff(Operator):
-    Input = InputSlot() # The last two non-singleton axes (except 'c') are the axes of the 'pages'.
-                        # Re-order the axes yourself if you want an alternative slicing direction
+    """Export to multi-page OME-TIFF.
+
+    Attributes:
+        Input: Image data source (input slot).
+        Filepath: Path to the exported image (input slot).
+        progressSignal: Subscribe to this signal to receive export progress updates.
+    """
+
+    Input = InputSlot()
     Filepath = InputSlot()
 
-    DEFAULT_BATCH_SIZE = 4
+    _DEFAULT_BATCH_SIZE = 4
+    # OME-TIFF requires 5D with arbitrary order.
+    _EXPORT_AXES = "tzcyx"
 
     def __init__(self, *args, **kwargs):
-        super(OpExportMultipageTiff, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
+
         self.progressSignal = OrderedSignal()
+
         self._opReorderAxes = OpReorderAxes(parent=self)
-        self._opReorderAxes.Input.connect( self.Input )
+        self._opReorderAxes.Input.connect(self.Input)
+        self._opReorderAxes.AxisOrder.setValue(self._EXPORT_AXES)
+
+        self._page_buf = None
 
     def setupOutputs(self):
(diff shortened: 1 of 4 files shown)


lazyflow/operators/ioOperators/opExportMultipageTiff.py
```diff
@@ -17,246 +17,265 @@
...
-        self.progressSignal(100)
+    def propagateDirty(self, slot, subindex, roi):
+        pass
 
-    # No output slots...
-    def execute(self, slot, subindex, roi, result): pass 
-    def propagateDirty(self, slot, subindex, roi): pass
+    def run_export(self) -> None:
+        """Export an image from Input to Filepath."""
+        path = pathlib.Path(self.Filepath.value)
```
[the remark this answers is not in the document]
emilmelnikov: I've used `pathlib` instead of `os` for something else, and just didn't want to `import os`, and also it's API is nicer and less error-prone. In this case it doesn't really matter, but I think it's better to use `pathlib` in new code anyway for consistency.

k-dominik:
@emilmelnikov I assume that in order to profit from the potential speedups, the "export everything to n5 first" needs to be implemented, right?

btw. do you have any idea why this is not run appveyor?

emilmelnikov:
> @emilmelnikov I assume that in order to profit from the potential speedups, the "export everything to n5 first" needs to be implemented, right?

Well, right now this PR should give us a modest speedup (and also somewhat refactored code), but for ~2x speedups we need N5 first. Alternatively, we can think about our own custom temporary on-disk cache format (pros: potentially more efficient compression/layout/etc., cons: not usable as-is for users).

> btw. do you have any idea why this is not run appveyor?

No idea, that's strange...

k-dominik:
btw, since your on it, do you think someone could be added to the test?
Otherwise LGTM

emilmelnikov:
> btw, since your on it, do you think someone could be added to the test?

Do you mean unit testing? There is already a unit test, and it passes (the API didn't change).

k-dominik:
I know that there are unit tests and that they pass. The question is have
you looked at them and do you find them adequate

btw, since your on it, do you think someone could be added to the test?
>
> Do you mean unit testing? There is already a unit test, and it passes (the
> API didn't change
>

emilmelnikov:
I think the test was OK, but I've (hopefully) made it more readable and 
"pytest-y".

tests/testOpExportMultipageTiff.py
```diff
@@ -18,96 +17,71 @@
...
-from lazyflow.operators.opArrayPiper import OpArrayPiper
+from lazyflow.graph import Graph
+from lazyflow.operator import Operator
 from lazyflow.operators.ioOperators import OpExportMultipageTiff
 from lazyflow.operators.ioOperators.opTiffReader import OpTiffReader
+from lazyflow.operators.opArrayPiper import OpArrayPiper
+from lazyflow.operators.opReorderAxes import OpReorderAxes
+
+
+class _Pipeline:
```
k-dominik: this is a very elegant pattern :+1:

tests/testOpExportMultipageTiff.py
```diff
@@ -18,96 +17,71 @@
...
-                opReorderAxes.cleanUp()
-        finally:
-            opReader.cleanUp()
+    graph = Graph()
 
-if __name__ == "__main__":
-    # Run this file independently to see debug output.
-    logger.setLevel(logging.DEBUG)
-    logger.addHandler(logging.StreamHandler(sys.stdout))
+    with tempfile.NamedTemporaryFile(suffix=".tiff") as file:
```
k-dominik: this fails on windows :/

from [the docs](https://docs.python.org/3.6/library/tempfile.html#tempfile.NamedTemporaryFile):
> This function operates exactly as TemporaryFile() does, except that the file is guaranteed to have a visible name in the file system (on Unix, the directory entry is not unlinked). That name can be retrieved from the name attribute of the returned file-like object. **Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later)**. If delete is true (the default), the file is deleted as soon as it is closed. The returned object is always a file-like object whose file attribute is the underlying true file object. This file-like object can be used in a with statement, just like a normal file.

emilmelnikov: Oops, missed it, sorry. Replaced it with the temporary directory (could use `mkstemp`, but contextmanagers are too awesome).

emilmelnikov:
> This is a really nice PR. I have encountered one thing, though, when exporting a 3D time series: First of all, the behavior is the same as before your changes, so it's nothing you have introduced. But still, when exporting it from ilastik, (I've used the Drosophila3D+t) I get one volume that is interpreted to have 2500 z-slices by Fiji. Do you think this could be resolved by writing something to the header?

This is an issue with the Fiji's standard file opener. The resulting TIFF opens correctly if you use `File -> Import -> BioFormats`.

k-dominik:
thanks @emilmelnikov :))

PR #293, opened 2019-02-28
Adds Input and AxisOrder to OpReorderAxes' constructor

Tomaz-Vieira:
Creates a more sensible constructor for OpReorderAxes, so that you can spawn it already configured and ready to go.

This is the first step towards having sensible operator constructors in Lazyflow, so that we can actually spawn operators that are ready to do their job and that, hopefully, are never inconsistent. This would eventually mean that `operator.configured()` would always be `True`. This change is convenient for the code cleanups that are coming in the object classification workflow, but those are in the ilastik project.

Right now operator construction is meaningless; one can do `OpReorderAxes()` and receive back an operator with dangling inputs. In fact, the programmer can even forget to set up those inputs, and be faced with the most common and most annoying exception in ilastik: `SlotNotReady`. Furthermore, doing this:
```
op5 = OpReorderAxes()
op5.AxisOrder = 'xyz'
op5.Input.connect(opWhatever.Output)
```
is far more inconvenient and error prone than simply

```
op5 = OpReorderAxes(AxisOrder='xyz', Input=opWhatever.Output)
```

not to mention that the first style can have more unpredictable `setupOutputs()` firings and somewhat violates the operator's encapsulation.

The idea is to eventually have operators that take all of their inputs in the constructor, so that it is instantiated in a consistent state, and then only allow the state to be modified in a transaction-like style (or maybe not allow it to be modified at all). Though this could be potentially done in the root `Operator` class by scanning `**kwargs` for operator names, I'd rather have the inputs explicitly defined in the constructor arguments so that the code is easier to follow (one can immediately see what can be passed into the operator and we can have type hints for every parameter).

### DIFF
--- a/lazyflow/operators/opReorderAxes.py
+++ b/lazyflow/operators/opReorderAxes.py
@@ -36,15 +36,13 @@ logger = logging.getLogger(__name__)
 
 class OpReorderAxes(Operator):
     Input = InputSlot()
-    AxisOrder = InputSlot(value='tzyxc') # string: The desired output axis order
+    AxisOrder = InputSlot() # string: The desired output axis order
     Output = OutputSlot()
 
-    def __init__(self, graph=None, parent=None, Input=None, AxisOrder=None):
+    def __init__(self, graph=None, parent=None, Input=None, AxisOrder='tzyxc'):
         super().__init__(graph=graph, parent=parent)
-        if Input is not None:
-            self.Input.connect(Input)
-        if AxisOrder is not None:
-            self.AxisOrder.setValue(AxisOrder)
+        self.Input.setOrConnectIfAvailable(Input)
+        self.AxisOrder.setOrConnectIfAvailable(AxisOrder)
 
     def setupOutputs(self):
         if 'c' not in self.AxisOrder.value:
--- a/lazyflow/operators/opReorderAxes.py
+++ b/lazyflow/operators/opReorderAxes.py
@@ -39,6 +39,13 @@ class OpReorderAxes(Operator):
     AxisOrder = InputSlot(value='tzyxc') # string: The desired output axis order
     Output = OutputSlot()
 
+    def __init__(self, graph=None, parent=None, Input=None, AxisOrder=None):
+        super().__init__(graph=graph, parent=parent)
+        if Input is not None:
+            self.Input.connect(Input)
+        if AxisOrder is not None:
+            self.AxisOrder.setValue(AxisOrder)
+
     def setupOutputs(self):
         if 'c' not in self.AxisOrder.value:
             # This is helpful in order to convert our internal axis order to 'tczyx'
--- a/lazyflow/slot.py
+++ b/lazyflow/slot.py
@@ -483,6 +483,17 @@ class Slot(object):
             self.meta._ready = False
             self._sig_unready(self)
 
+    def setOrConnect(self, value_or_slot):
+        if isinstance(value_or_slot, Slot):
+            self.connect(value_or_slot)
+        else:
+            self.setValue(value_or_slot)
+
+    def setOrConnectIfAvailable(self, value_or_slot):
+        if value_or_slot is None:
+            return
+        self.setOrConnect(value_or_slot)
+
     @is_setup_fn
     def connect(self, upstream_slot, notify=True, permit_distant_connection=False):
         """
--- a/tests/testOpReorderAxes.py
+++ b/tests/testOpReorderAxes.py
@@ -60,27 +60,26 @@ class TestOpReorderAxes(unittest.TestCase):
         self.array = None
         self.axis = list('tzyxc')
         self.tests = 20
-        graph = Graph()
-        self.operator = OpReorderAxes(graph=graph)
+        self.graph = Graph()
+        self.operator = OpReorderAxes(graph=self.graph)
 
-    def prepareVolnOp(self, possible_axes='tzyxc', num=5):
-        tags = random.sample(possible_axes,random.randint(2,num))
-        tagStr = ''
-        for s in tags:
-            tagStr += s
+    def prepareVolnOp(self, possible_axes='tzyxc', num=5, AxisOrder=None, config_via_init=False):
+        tagStr = ''.join(random.sample(possible_axes,random.randint(2,num)))
         axisTags = vigra.defaultAxistags(tagStr)
 
-        self.shape = []
-        for tag in axisTags:
-            self.shape.append(random.randint(20,30))
+        self.shape = [random.randint(20,30) for tag in axisTags]
 
-        self.array = (numpy.random.rand(*tuple(self.shape))*255)
+        self.array = (numpy.random.rand(*self.shape)*255)
         self.array =  (float(250)/255*self.array + 5).astype(int)
         self.inArray = vigra.VigraArray(self.array,axistags = axisTags)
 
-        opProvider = OpArrayProvider( graph=self.operator.graph )
+        opProvider = OpArrayProvider( graph=self.graph )
         opProvider.Input.setValue( self.inArray )
-        self.operator.Input.connect( opProvider.Output )
+
+        if config_via_init:
+            self.operator = OpReorderAxes(graph=self.graph, Input=opProvider.Output, AxisOrder=AxisOrder)
+        else:
+            self.operator.Input.connect(opProvider.Output)
 
     def test_Full(self):
         for i in range(self.tests):
@@ -156,12 +155,14 @@ class TestOpReorderAxes(unittest.TestCase):
         
     def _impl_roi_custom_order(self, axisorder):
         for i in range(self.tests):
-            self.prepareVolnOp(axisorder, len(axisorder)-1)
-            
+            use_constructor = bool(i%2)
             # Specify a strange order for the output axis tags
-            self.operator.AxisOrder.setValue(axisorder)
+            self.prepareVolnOp(axisorder, len(axisorder)-1, AxisOrder=axisorder, config_via_init=use_constructor)
+            if not use_constructor:
+                self.operator.AxisOrder.setValue(axisorder)
+
             shape = self.operator.Output.meta.shape
-            
+
             roi = [None,None]
             roi[1]=[numpy.random.randint(2,s) if s != 1 else 1 for s in shape]
             roi[0]=[numpy.random.randint(0,roi[1][i]) if s != 1 else 0 for i,s in enumerate(shape)]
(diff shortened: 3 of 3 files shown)


lazyflow/operators/opReorderAxes.py
```diff
@@ -39,6 +39,13 @@ class OpReorderAxes(Operator):
     AxisOrder = InputSlot(value='tzyxc') # string: The desired output axis order
     Output = OutputSlot()
 
+    def __init__(self, graph=None, parent=None, Input=None, AxisOrder=None):
```
k-dominik: So input should be a slot and axisorder a string then. What about connecting the output of a different operator that spits out an axiorder?

Tomaz-Vieira:
> Could you maybe add a rational, what was lacking before, what problem does this solve? (I mean is there some code that would benefit from this?)
> This operator already has some tests, so it should be trivial to add one to this as well.
> 
> On a more bigger picture view: What is so special about `OpReorderAxes` that it gets this treatment? Should we maybe think of adding `**kwargs` to the base class and enable this kind of behavior in all operators?

I'll add

lazyflow/operators/opReorderAxes.py
Tomaz-Vieira: True. I guess I should be accepting slots as well as strings here

emilmelnikov:
I agree that construction initialization is better, but maybe we can somehow make this change globally in terms of API design? That is, make all operators receive all their inputs in `__init__`.

Tomaz-Vieira:
> I agree that construction initialization is better, but maybe we can somehow make this change globally in terms of API design? That is, make all operators receive all their inputs in `__init__`.
I think we should change it everywhere we can, but it won't be a simple job. But we gotta start somewhere =)

emilmelnikov:
We can modify `Operator` and connect inputs in `__init__`:

```python
class Operator(...):
    def __init__(self, ..., inputs: Optional[Mapping[str, Any]] = None):
        if inputs is None:
            inputs = {}
        for input_name, output in inputs.items():
            if isinstance(output, Slot):
                self.inputs[input_name].connect(output)
            else:
                self.inputs[input_name].setvalue(output)

...

other_op = SomeOp(...)
inputs = dict(Input=other_op.Input, AxisOrder="tczyx")
op = MyOp(..., inputs=inputs)
```

Tomaz-Vieira:
> We can modify `Operator` and connect inputs in `__init__`:
> 
> ```python
> class Operator(...):
>     def __init__(self, ..., inputs: Optional[Mapping[str, Any]] = None):
>         if inputs is None:
>             inputs = {}
>         for input_name, output in inputs.items():
>             if isinstance(output, Slot):
>                 self.inputs[input_name].connect(output)
>             else:
>                 self.inputs[input_name].setValue(output)
> 
> ...
> 
> other_op = SomeOp(...)
> inputs = dict(Input=other_op.Input, AxisOrder="tczyx")
> op = MyOp(..., inputs=inputs)
> ```

I'm still not sure what to think about moving this logic up to the base Operator class... It's cool that it would apply to all operators in one fell swoop, but it has some disadvantages:
- There's no control over the order in which the slots are set up, which might fire `setupOutputs` several times if the first slots to be set are the optional ones (unless we think of some transaction-like mechanism here, which might also be a good idea);
- Speaking of optional slots, having things like `SomeSlot=None` in the constructor makes it clear that it is indeed optional and produces better, immediate erros if one forgets to set it up (i.e.: `missing required argument SomeSlot...`);
- In order for a dev to know what he is supposed to provide to the constructor, he still has to look up the Operator's "static" Inputs, as well as the inputs for every one of its ancestors' operators;
- We can't add type annotations for every slot individually (though that would still be a problem if we're allowing both values or slots to be passed in);
- If a child operator needs to set the value of some slot in its specialized constructor, then it would either have to remove that slot from the dict passed up to it's parent, or not call the parent's `__init__` at all or maybe be ready for it to be set multiple times

I guess I'd prefer the explicit version, with all parameters in the constructor, rather than the implicit (though admittedly less boiler-plate-y ) one.

emilmelnikov:
> * There's no control over the order in which the slots are set up, which might fire `setupOutputs` several times if the first slots to be set are the non-optional ones (unless we think of some transaction-like mechanism here, which might also be a good idea);

There is an order if one uses an `OrderedDict` (or just a usual `dict` since CPython 3.6 / Python 3.7).

> * Speaking of optional slots, having things like `SomeSlot=None` in the constructor makes it clear that it is indeed optional and produces better, immediate erros if one forgets to set it up (i.e.: `missing required argument SomeSlot...`);

We already have optional slots, they just need to be properly documented and/or type-annotated, and sanity checks could also be implemented in the base `Operator` because we know which slots are optional and which are not.

> * In order for a dev to know what he is supposed to provide to the constructor, he still has to look up the Operator's "static" Inputs, as well as the inputs for every one of its ancestors' operators;

Since you can access them as instance attributes, slots could (should?) be documented in the `Attributes` class docstring section (for the Google style guide), and you need to consult the documentation anyway.

> * We can't add type annotations for every slot individually (though that would still be a problem if we're allowing both values or slots to be passed in);

Starting with Python 3.6 and [PEP 526](https://www.python.org/dev/peps/pep-0526/), we can annotate class-level slot variables.

> * If a child operator needs to set the value of some slot in its specialized constructor, then it would either have to remove that slot from the dict passed up to it's parent, or not call the parent's `__init__` at all or maybe be ready for it to be set multiple times

You should be able to assign a value to a slot in the child operator too, otherwise this breaks Liskov substitution principle. If you just want to initialize the slot with a different value, document it in the constructor (I suppose these cases are relatively rare).

> I guess I'd prefer the explicit version, with all parameters in the constructor, rather than the implicit (though admittedly less boiler-plate-y ) one.

The main question is: what operators do with their input slots? If they mostly connect them to some outputs and/or do other very generic stuff, then the generic version is better. Otherwise, the ad-hoc version makes more sense.

k-dominik:
> ```
> op5 = OpReorderAxes()
> op5.AxisOrder = 'xyz'
> op5.Input.connect(opWhatever.Output)
> ```
> 
> 
> is far more inconvenient and error prone than simply

inconvenient I get. Error-prone I don't.

There is not much to add to @emilmelnikov's answer. I also think this should be done at the operator level. Maybe assigning values during the construction should be a special case of a `setInputSlots` method that actually does a transaction. This would be way more useful than just having a constructor. Just think changing values in the export applet.


> > * There's no control over the order in which the slots are set up, which might fire `setupOutputs` several times if the first slots to be set are the non-optional ones (unless we think of some transaction-like mechanism here, which might also be a good idea);
> 
> There is an order if one uses an `OrderedDict` (or just a usual `dict` since CPython 3.6 / Python 3.7).

I think if we wanted it would be quite easy to setup a transaction mechanism (which would be helpful in any case). Just think of adding a `_transaction_slot = InputSlot(value=True)`.

And then,

```python
    # somewhere in the operator class
    @contextmanager
    def transaction(self):
        self._transaction_slot.setValue(False)
        try:
            yield None
        finally:
            self._transaction_slot.setValue(True)
```

this would allow

```python
    def setInputSlots(self, inputs: Optional[Mapping[str, Any]] = None):
        with self.transaction:
            # emil's code:
            if inputs is None:
                inputs = {}
                for input_name, output in inputs.items():
                if isinstance(output, Slot):
                    self.inputs[input_name].connect(output)
                else:
                    self.inputs[input_name].setValue(output)
```

then you could at least

```python
op5 = OpReorderAxes(parent=...)
op5.setInputSlots(inputs={'axesorder': xyz', 'Input': opWhatever.Output})
```

the only thing a bit scary is an operator setting it's own input....

Tomaz-Vieira:
> inconvenient I get. Error-prone I don't.

Error prone because:
- one can forget to set one of the slots;
- one can set up the slots in the wrong order;
- one can set up later slots in a way that is inconsistent with earlier slots;
- It allows for an operator to be setup across several places in the codebase, making it hard to track where it's getting its inputs.

This probably wouldn't be the case for `OpReorderAxes`, but could definitely be for the ops that have a dozen slots. Also, the constructor style makes it so that you can trust that your operator has been properly configured whenever you have an operator instance in whatever context you could possibly be. Think of what a method like this means If we're able to spawn operators in an inconsistent state:
```
def my_method(self, some_op:OpMyOperator):
    bla = some_op.some_method(123) #is some_op ready to respond to some_method?
    ...

```
It really doesn't matter much that `my_method` knows that it's getting an instance of `OpMyOperator`, becase it can't trust that instance to do anything since it might not be initialized. The main point is: if the operator needs some stuff to get going, why is it that we don't provide it with all of that stuff upon creation? If we're not gonna initialize the objects in the `__init__` function, than we might just as well be using dictionaries everywhere.

 >> There is an order if one uses an `OrderedDict` (or just a usual `dict` since CPython 3.6 / Python 3.7).

Yeah, but are you, the client of the Op* class, supposed to know which is the correct order? Should the constructor throw an exception telling you that you had the wrong order and that you should use order "SlotA, SlotC, SlotD" instead? Should that be hardcoded somewhere? We can invent a bunch of mechanisms to replace python's keyword arguments, but why not stick to the tried and true convention, which is bug-free and easy for outsiders to understand?

>We already have optional slots, they just need to be properly documented and/or type-annotated, and sanity checks could also be implemented in the base Operator because we know which slots are optional and which are not.

Yes, we can do all that, but Python could be doing that for us, in a much more readable/conventional way and at essentially no development or maintenance cost.

>There is not much to add to @emilmelnikov's answer. I also think this should be done at the operator level. Maybe assigning values during the construction should be a special case of a setInputSlots method that actually does a transaction. This would be way more useful than just having a constructor. Just think changing values in the export applet

I think that in many situations it's best to discard old operators and build them anew than to try to mutate it into the new configuration that you want. I'm not sure if that would be viable in ilastik, though, but even if it's not, i think it would be fine to have a `setupInputs` that takes all the necessary inputs and sets up the operator in a single transaction.

> I think if we wanted it would be quite easy to setup a transaction mechanism (which would be helpful in any case). Just think of adding a `_transaction_slot = InputSlot(value=True)`.

I'm not really comfortable with this approach... it abuses the knowledge that setupOutputs won't trigger unless all Inputs are connected and it creates an input which doesn't really provide input to anything (nothing will ever read from that input) and, as you said, it requires the operator to set up its own input, which feels wrong ^^

k-dominik:
heyo @Tomaz-Vieira could you please add the handling of slots vs. values to this PR? (just if axisorder is a slot as well, it should be connected)

Tomaz-Vieira:
> heyo @Tomaz-Vieira could you please add the handling of slots vs. values to this PR? (just if axisorder is a slot as well, it should be connected)

Sure. I think I already had a commit for this but forgot to push. Coming right up.

PR #294, opened 2019-03-05
merge tiktorchLazyflowClassifier and tiktorchLazyflowFactory

FynnBe:
Pave the way for 'LazyflowOnlineClassifier'...
only makes sense with ilastik/ilastik@477e2082435d2ed7c3cb530ad39bf1c99e5689f7

### DIFF
--- a/lazyflow/classifiers/lazyflowClassifier.py
+++ b/lazyflow/classifiers/lazyflowClassifier.py
@@ -2,11 +2,14 @@ from builtins import object
 import abc
 from future.utils import with_metaclass
 
-def _has_attribute( cls, attr ):
+
+def _has_attribute(cls, attr):
     return any(attr in B.__dict__ for B in cls.__mro__)
 
-def _has_attributes( cls, attrs ):
+
+def _has_attributes(cls, attrs):
     return all(_has_attribute(cls, a) for a in attrs)
+
 
 class LazyflowVectorwiseClassifierFactoryABC(with_metaclass(abc.ABCMeta, object)):
     """
@@ -52,10 +55,11 @@ class LazyflowVectorwiseClassifierFactoryABC(with_metaclass(abc.ABCMeta, object)
         """
         if cls is LazyflowVectorwiseClassifierFactoryABC:
             is_subclass = LazyflowVectorwiseClassifierFactoryABC in C.__mro__
-            is_subclass &= _has_attributes(C, ['create_and_train', 'description'])
-            is_subclass &= 'VERSION' in C.__dict__
+            is_subclass &= _has_attributes(C, ["create_and_train", "description"])
+            is_subclass &= "VERSION" in C.__dict__
             return is_subclass
         return NotImplemented
+
 
 class LazyflowVectorwiseClassifierABC(with_metaclass(abc.ABCMeta, object)):
     """
@@ -100,7 +104,7 @@ class LazyflowVectorwiseClassifierABC(with_metaclass(abc.ABCMeta, object)):
     @classmethod
     def __subclasshook__(cls, C):
         if cls is LazyflowVectorwiseClassifierABC:
-            return _has_attributes(C, ['predict_probabilities', 'known_classes', 'serialize_hdf5', 'deserialize_hdf5'])
+            return _has_attributes(C, ["predict_probabilities", "known_classes", "serialize_hdf5", "deserialize_hdf5"])
         return NotImplemented
 
     @abc.abstractmethod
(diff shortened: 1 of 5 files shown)


lazyflow/classifiers/tiktorchLazyflowClassifier.py
```diff
@@ -349,17 +297,18 @@ def predict_probabilities_pixelwise(self, feature_image, roi, axistags=None):
...
-        self._opReorderAxesOut.AxisOrder.setValue(''.join(axistags.keys()))
-        self._opReorderAxesOut.Input.setValue(vigra.VigraArray(result, axistags=vigra.defaultAxistags('czyx')))
+        self._opReorderAxesOut.AxisOrder.setValue("".join(axistags.keys()))
+        self._opReorderAxesOut.Input.setValue(vigra.VigraArray(result, axistags=vigra.defaultAxistags("czyx")))
         return self._opReorderAxesOut.Output[:].wait()
 
     @property
     def known_classes(self):
-        nr_classes = self._config.get('output_shape')[0]
+        nr_classes = self._config.get("output_shape")[0]
```
emilmelnikov: This will fail if "output_shape" is not set because `None` cannot be indexed. If this is the intended behavior, you can just use an ordinary indexing instead of `get`.

Or did you mean `get("output_shape", 0)`?

lazyflow/classifiers/tiktorchLazyflowClassifier.py
```diff
@@ -60,44 +56,60 @@ def __init__(
...
-            laucher = RemoteSSHServerLauncher(
-                conn_conf,
-                cred=SSHCred(server_config['username'], server_config['password'])
+            self.launcher = RemoteSSHServerLauncher(
+                conn_conf, cred=SSHCred(server_config["username"], server_config["password"])
             )
 
-        laucher.start()
+        self.launcher.start()
+        self._shutdown_sent = False
```
emilmelnikov: Is it possible to ask Launchers about their status and remove the `_shutdown_sent` variable?

lazyflow/classifiers/lazyflowClassifier.py
```diff
@@ -260,12 +260,166 @@ def serialize_hdf5(self, h5py_group):
...
+
+    @abc.abstractmethod
+    def get_halo_shape(self, data_axes="zyxc"):
+        """
+        Same as LazyflowPixelwiseClassifierFactoryABC.get_halo_shape().  See that function for details.
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def __subclasshook__(cls, C):
```
emilmelnikov: Are subclasshooks really necessary in these cases? Seems like a design problem...

lazyflow/classifiers/tiktorchLazyflowClassifier.py
```diff
@@ -113,51 +125,40 @@ def pause_training_process(self):
...
-            reordered_feature_images.append(self._opReorderAxes.Output([]).wait())
+            self._opReorderAxesIn.Input.setValue(vigra.VigraArray(feature_images[i], axistags=axistags))
+            reordered_feature_images.append(self._opReorderAxesIn.Output([]).wait())
 
-            self._opReorderAxes.Input.setValue(
-                vigra.VigraArray(label_images[i], axistags=axistags)
-            )
-            self._opReorderAxes.AxisOrder.setValue('czyx')
-            reordered_labels.append(self._opReorderAxes.Output([]).wait())
+            self._opReorderAxesIn.Input.setValue(vigra.VigraArray(label_images[i], axistags=axistags))
```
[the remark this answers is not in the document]
m-novikov: I suppose in this case we are using `self._opReorderAxesIn` as a function (to reorder axes). Intent obscured by lazyflow.

FynnBe:
Thanks @emilmelnikov for reviewing these changes. That is some great input! However, this is work in progress on the alpha dev branch... I merely made this premature PR to communicate my current changes with @m-novikov . Right now this is basically just copy-pasted from the LazyflowPixelwiseClassifier.

lazyflow/classifiers/tiktorchLazyflowClassifier.py
```diff
@@ -60,44 +56,60 @@ def __init__(
...
+        self.launcher.start()
+        self._shutdown_sent = False
+
+        self._tikTorchClient = Client(INeuralNetworkAPI(), conn_conf)
+        self._tikTorchClient.load_model(config, binary_model, binary_state, binary_optimizer_state)
 
-        self._config = config
-        self.tikTorchClient = Client(INeuralNetworkAPI(), conn_conf)
-        self.tikTorchClient.load_model(config, binary_model, binary_state, binary_optimizer_state)
+    def __getattr__(self, item):
```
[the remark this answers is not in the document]
m-novikov: That's not true. `__getattr__` is only called if standard lookup mechanism failed (from `__dict__`).
I think you're mixing this with `__getattribute__`, but in this case this code wouldn't work at all.
See example below:
```pyhon
def test_attr():
    undef = object()
    class Foo:
        def __init__(self):
            self._config = {'a': 1}

        def __getattr__(self, name):
            val = self._config.get(name, undef)
            if val is undef:
                raise AttributeError(name)
            return val

    f = Foo()

    assert f.a == 1
    assert f._config == {'a': 1}
```

PR #296, opened 2019-04-05
use new deps in ci

k-dominik:


appveyor.yml
```diff
@@ -6,11 +6,10 @@ environment:
   MINICONDA: C:\Miniconda36-x64
```
m-novikov: Except the name is still 3.6

k-dominik: doesn't really matter - conda will install the correct python version bc. of the depencies - but i'll change to make it more explicit!

PR #297, opened 2019-04-08
Request tests

m-novikov:


Tomaz-Vieira:
It's great that we have thse tests now =)

My paranoia tells me that it would be good to also test that Request notifications (say, `notify_failed`) also fire even if registered _after_ the `Request` has been submitted and even completed. It seems to me that creating a `Request` object should automatically submit it to the threadpool, so if we ever get around to fixing that (I tried and failed because of the `writeInto` method), it would be great to make sure the notification mechanisms still work.

Still, I'd support this PR even without that.

tests/test_request/test_request.py
```diff
@@ -0,0 +1,276 @@
...
+            assert req.wait() == 42
+
+        req.notify_failed(recv)
+        recv.assert_called_once()
+        assert isinstance(recv.call_args[0][0], TExc)
+
+
+@pytest.fixture
+def work():
+    cont = threading.Event()
```
k-dominik: could be at least something like `continue` or `unpause`
```suggestion
    continue = threading.Event()
```

tests/test_request/test_threadPool.py
```diff
@@ -0,0 +1,122 @@
...
+
+    pool.wake_up(task)
+    assert start.wait(timeout=1)
+    assert pool.get_states().count("running task") == 1
+    assert stop.wait(timeout=1)
+
+
+def test_wake_task_executes_task_on_assigned_worker(pool: ThreadPool):
+    stop = threading.Event()
+    box = [None]
```
k-dominik: could use `nonlocal`

tests/test_request/test_threadPool.py
```diff
@@ -0,0 +1,122 @@
...
+        time.sleep(0.2)
+        stop.set()
+
+    task.assigned_worker = random.choice(list(pool.workers))
+    pool.wake_up(task)
+    assert stop.wait(timeout=1)
+    assert box[0] == task.assigned_worker
+
+
+def test_exception_does_not_kill_worker(caplog):
```
k-dominik: `caplog` not used?!

tests/test_request/test_request.py
```diff
@@ -0,0 +1,276 @@
...
+
+        signal.subscribe(f)
+
+        signal(1, 2, 3, val=2)
+
+        f.assert_called_with(1, 2, 3, val=2)
+
+    def test_subscribed_functions_called_in_sequence(self, signal: SimpleSignal):
+        order = []
+        for c in range(100):
```
emilmelnikov: This could work without a custom func:
```python
for c in range(100):
    signal.subscribe(partial(order.append, c))
```

tests/test_request/test_request.py
```diff
@@ -0,0 +1,276 @@
...
+
+            def fun(val=c):
+                order.append(val)
+
+            signal.subscribe(fun)
+
+        signal()
+        assert len(order) == 100
+
+        for idx in range(len(order) - 1):
```
emilmelnikov:
```python
assert order == sorted(order)
```

tests/test_request/test_request.py
```diff
@@ -0,0 +1,276 @@
...
+        finally:
+            self.done.set()
+
+
+class TestRequest:
+    def test_submit_should_assign_worker_and_execute(self):
+        stop = threading.Event()
+
+        def work():
+            time.sleep(0.3)
```
emilmelnikov: Why sleep here, and why `stop` is needed?

tests/test_request/test_request.py
```diff
@@ -0,0 +1,276 @@
...
+
+def test_cancels_child_requests(cancelled_work):
+    assert len(cancelled_work.children) == 2
+
+    for ch in cancelled_work.children:
+        assert ch.cancelled
+
+    work_rq = cancelled_work.request
+
+    assert work_rq.child_requests == set()
```
emilmelnikov:
```suggestion
    assert not work_rq.child_requests
```

Unless `child_requests` could be something else than `set`.

tests/test_request/test_threadPool.py
```diff
@@ -0,0 +1,122 @@
...
+class Task:
+    _counter = itertools.count()
+
+    def __init__(self, fn):
+        self.fn = fn
+        self.assigned_worker = None
+        self.priority = next(self._counter)
+
+    def __lt__(self, other):
+        return self.priority > self.priority
```
emilmelnikov:
```suggestion
        return self.priority < other.priority
```

tests/test_request/test_threadPool.py
```diff
@@ -0,0 +1,122 @@
...
+        self.fn()
+
+
+@pytest.fixture
+def pool():
+    return ThreadPool(NUM_WORKERS)
+
+
+def test_thread_pool_starts_workers(pool: ThreadPool):
+    assert pool.num_workers == NUM_WORKERS
```
emilmelnikov:
```python
assert pool.num_workers == len(pool.workers) == NUM_WORKERS
```

tests/test_request/test_threadPool.py
```diff
@@ -0,0 +1,122 @@
...
+    assert len(pool.workers) == NUM_WORKERS
+
+    for w in pool.workers:
```
emilmelnikov:
```python
assert all(isinstance(w, threading.Thread) for w in pool.workers))
```

m-novikov: There is a problem to consider with failing test:
```
E        +  where False = all(<generator object test_foo.<locals>.<genexpr> at 0x7ff77d0ca7c8>)
```
While current form shows exact offending object without further investigation

PR #303, opened 2019-06-11
Operator transaction

m-novikov:


lazyflow/slot.py
```diff
@@ -1479,15 +1479,22 @@ def _changed(self):
...
         are connected and configured.
 
         """
         if self.operator is not None:
             # check whether all slots are connected and notify operator
-            if self.operator.configured():
-                self.operator._setupOutputs()
+            transaction = getattr(self.operator, "_current_transaction", None)
+            if transaction:
+                transaction.on_exit(self._invokeSetupOutputs)
```
Tomaz-Vieira: I think that we have some code like this in some of the workflows:
```
op1 = SomeOperator(...)
op2 = SomeOtherOperator(...)
op2.SomeOtherInput.connect(op1.SomeInput)
```
that is, the upstream slot doesn't have to be an OutputSlot. If you, then, do this:`
```
with op1.transaction:
    op1.SomeInput.setValue(123)
```
`op2` there would detect the change in `SomeOtherInput` and call its own `setupOutpus`. You might want to double check my reasoning here, though =)

lazyflow/operator.py
```diff
@@ -182,6 +203,17 @@ class Operator(metaclass=OperatorMetaClass):
     description = ""
     category = "lazyflow"
 
+    _current_transaction = None
```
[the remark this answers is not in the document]
m-novikov: This is supposed to work this way:
1. If there is no initialized transaction on class, then `self._current_transaction` will be retrieved as a class variable
2. Otherwise, we will set an instance variable on the operator, which will shadow the class level variable
Idea was to have a universal interface and avoid `getattr` calls checking conformity to this interface (operator always reliably provide `_current_transaction`)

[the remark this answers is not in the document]
Tomaz-Vieira: Wow, I didn't know it would get shadowed rather than overwritten o.O
But is there any Operator that doesn't ultimately inherit from the `Operator` base class that would justify not having `self._current_transaction = None` in Operator's `__init__`? Or in `__new__`, for that matter, since we're already doing that crazy slot replication magic anyway...

lazyflow/slot.py
m-novikov: Updated, see [this test case](https://github.com/ilastik/lazyflow/pull/303/files#diff-5eb200b7b0b313e34a894843a84f0a14R734)

lazyflow/operator.py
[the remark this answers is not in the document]
m-novikov: It's more common to have this type of default values on class level and if we at some point will have type annotations they will be also on the class level so we'll anyway have this variable there.

emilmelnikov:
I am a bit confused with the fact that this feature is called "transaction". I would expect the following behaviour:

```python
op.Input.setValue("spam")
with op.transaction:
    op.Input.setValue("eggs")
    assert op.Slot1.value == "spam"
assert op.Input.value == "eggs"
```

lazyflow/operator.py
```diff
@@ -30,6 +30,28 @@
 from lazyflow.slot import InputSlot, OutputSlot, Slot
 
 
+class _Transaction:
```
emilmelnikov: This class could probably be moved into `Operator`:
```python
class Operator:
    def __init__(self):
        self._deferred_callbacks = None

    @contextmanager
    def transaction(self):
        assert self._deferred_callbacks is None
        self._deferred_callbacks = []
        yield
        callbacks = self._deferred_callbacks
        self._deferred_callbacks = None
        for cb in callbacks:
            cb()


class Slot:
    def _maybe_call_within_transaction(self, fn):
        op = getattr(self, "operator", None)
        callbacks = getattr(op, "_deferred_callbacks", None)
        if callbacks is None:
            fn()
        elif fn not in callbacks:
            callbacks.append(fn)
```

m-novikov:
> 
> 
> I am a bit confused with the fact that this feature is called "transaction". I would expect the following behaviour:

I am open to alternative naming. Currently when we speak about this "feature" we use `transaction slots` concept.

lazyflow/operator.py
m-novikov: I will move this class under Operator namespace, but I am reluctant putting yet another responsibility fully inside `Operator`.
> ```python
>         if callbacks is None:
>             fn()
>         elif fn not in callbacks:
>             callbacks.append(fn)
> ```
deduplication of callbacks done inside transaction by set

emilmelnikov: > I will move this class under Operator namespace, but I am reluctant putting yet another responsibility fully inside `Operator`.

Agreed. I suggested the change because `_Transaction` is small and mutually depends on `Operator`.

> deduplication of callbacks done inside transaction by set

This suggestion was made to preserve callback execution order: `set` doesn't do that, and `list` is simpler/faster than `dict` on small sizes.

lazyflow/operator.py
```diff
@@ -182,6 +203,17 @@ class Operator(metaclass=OperatorMetaClass):
...
+    _current_transaction = None
+
+    @property
+    def transaction(self):
+        """
+        Create transaction for this operation deferring setupOutputs call
+        until transaction is finished
+        :returns: Transaction context manager
+        """
+        return _Transaction(self)
```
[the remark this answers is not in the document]
k-dominik: i second the notion to return `self._current_transaction or _Transaction(self)`

k-dominik:
> I am a bit confused with the fact that this feature is called "transaction". I would expect the following behaviour:
> 
> ```python
> op.Input.setValue("spam")
> with op.transaction:
>     op.Input.setValue("eggs")
>     assert op.Input.value == "spam"
> assert op.Input.value == "eggs"
> ```

it's true that you kind of have to know our code in order to understand the meaning of transaction here. However, it is not a slot level transaction, but an operator level one, so I think this is fine if properly documented.

emilmelnikov:
> I think this is fine if properly documented.

LGTM

Also, I would advise against using `@property` for `transaction` because it visibly changes the instance state.

m-novikov:
> Also, I would advise against using `@property` for `transaction` because it visibly changes the instance state.

It doesn't change *public interface* only internal state. As an example, descriptors fields and cached properties, that also change internal state and still are properties.

m-novikov:
To make transactions work with `OperatorWrapper` I moved it to a graph level.

lazyflow/graph.py
```diff
@@ -80,16 +79,42 @@
...
+            return self._deferred_callbacks is not None
+
+        def on_exit(self, fn):
+            if fn in self._deferred_callbacks:
+                return
+            else:
+                self._deferred_callbacks.append(fn)
+
+        def __enter__(self):
+            assert self._deferred_callbacks is None, "Nested transactions are not supported"
```
k-dominik: wouldn't it be better to check `active` here?

lazyflow/graph.py
```diff
@@ -80,16 +79,43 @@
...
 
 
-class Graph(object):
+class Graph:
     """
     A Graph instance is shared by all connected operators and contains any
     bookkeeping or globally accessible state needed by all operators/slots in the graph.
     """
 
+    class Transaction:
```
emilmelnikov: Maybe we could support nested transactions just by keeping a list of lists?

```python
def __init__(self):
    self._deferred_callbacks = []

def on_exit(self, fn):
    assert self._deferred_callbacks
    cbs = self._deferred_callbacks[-1]
    if fn not in cbs:
        cbs.append(fn)        

def __enter__(self):
    self._deferred_callbacks.append([])

def __exit__(self, ...):
    for cb in self._deferred_callbacks.pop():
        cb()
```

k-dominik: Do you have a use-case in mind for nested transactions?

emilmelnikov: > Do you have a use-case in mind for nested transactions?

More composable functions/methods:

```python
with op.transaction:
    op.setValue("spam")
    do_stuff(op)

# 1000 lines later...

def do_stuff(op):
    with op.transaction:
        op.setValue("eggs")
```

Although I am not sure that this would be a common use case. However, this seems to be a low-hanging fruit.

m-novikov: I would defer implementing nested transactions.
For example what do you except on exit from a nested transaction? Should I be just reentrant transaction on the graph?

PR #305, opened 2019-06-13
Sane tracebacks

m-novikov:
Added `Operator Stack` section to exceptions see examples [here](https://gist.github.com/m-novikov/6e6d3fb7d4fdb55c2ec0b5377846927d)
The section includes full operator class name + method and pointer to file:lineno

### DIFF
--- a/lazyflow/operator.py
+++ b/lazyflow/operator.py
@@ -23,9 +23,11 @@ import collections
 import functools
 import logging
 import threading
+import sys
+import inspect
 
 from abc import ABCMeta
-from typing import Optional
+from traceback import walk_tb, FrameSummary, format_list
 
 # lazyflow
 from lazyflow.slot import InputSlot, OutputSlot, Slot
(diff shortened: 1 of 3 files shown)


emilmelnikov:
Marvelous!

Can we somehow attach the new, filtered operator traceback as a cause to the original one, to be consistent with the usual exception chaining concept?

m-novikov:
> Can we somehow attach the new, filtered operator traceback as a cause to the original one, to be consistent with the usual exception chaining concept?

Do you mean `raise e from <smth>`? In this case, I would like to avoid this, because it requiries reraising exception which is unconvinient to say the least in chained operators (basically one must only raise custom exception at the end of the operator chain) and doesn't provide nice output

lazyflow/operator.py
```diff
@@ -598,3 +600,60 @@ def debug_text(self):
...
+
+def print_operator_stack(exc_type, exc, tb):
+    """
+    Enrich default exception output with operator stacktrace
+    """
+    _original_excepthook(exc_type, exc, tb)
+
+    formatted = format_operator_stack(tb)
+    if formatted:
+        print("\n===Operator stack===\n")
```
k-dominik: would be cool to redirect that to stderr

k-dominik:
:+1: LGTM

lazyflow/operator.py
```diff
@@ -598,3 +600,60 @@ def debug_text(self):
...
+    _original_excepthook(exc_type, exc, tb)
+
+    formatted = format_operator_stack(tb)
+    if formatted:
+        print("\n===Operator stack===\n", file=sys.stderr)
+        print("".join(formatted), end="", file=sys.stderr)
+        print("\n===Operator stack end===\n", file=sys.stderr)
+
+
+sys.excepthook = print_operator_stack
```
emilmelnikov: Maybe make this feature opt-in? That is, provide `enable/disable_operator_stacktrace()` functions or something similar?
Then, we could conditionally call them in ilastik, maybe together with `--debug` switch.

m-novikov: I would like for end-users to send us these tracebacks. As long as we test it extensively don't see a reason to disable it. But I suppose we can make it opt-out?

emilmelnikov: I would rather get full tracebacks from users, so we can filter them later on our side.

By "opt-in" I meant not changing `sys.excepthook` when importing `lazyflow.operator`, but rather doing it explicitly somewhere in `ilastik.py`.

k-dominik: > I would rather get full tracebacks from users, so we can filter them later on our side.

this still produces the full traceback, it is called at the beginning of the function

m-novikov: > this still produces the full traceback, it is called at the beginning of the function
What I meant it would be nice to see this bit of information along with a traceback user sends us.

m-novikov:
Added context manager and install/uninstall functions.
Changed implementation a bit.
Now I inspect a `self` object to guess the Operator class, instead of looking it up within a module.
Seems like there is no sane way to get function `__qualname__` from frame or code object.
https://bugs.python.org/issue13672
https://bugs.python.org/issue12857

k-dominik:
:+1:

PR #308, opened 2019-06-21
Fix attribute error on opaque slots usage

m-novikov:


lazyflow/slot.py
```diff
@@ -931,26 +931,14 @@ def __call__(self, destination=None):
...
-                            " {} vs {}.  roi was {}".format(
-                                self.operator, result_op.shape, destination.shape, str(self.roi)
-                            )
-                        )
-                    destination = result_op
-
-                    # check that the returned value is compatible with the requested roi
-                    self.slot.stype.check_result_valid(self.roi, destination)
+                    if destination_given and id(result_op) != id(destination):
+                        # check that the returned value is compatible with the requested roi
```
k-dominik: isn't that what `slot.stype.check_result_valid` is supposed to do?

https://github.com/ilastik/lazyflow/blob/55858dd85af8c08be30fe4da018b8651cbeb7803/lazyflow/stype.py#L93-L98

tests/testOperatorInterface.py
```diff
@@ -764,6 +764,31 @@ def propagateDirty(self, *a, **kw):
...
+        def propagateDirty(self, *a, **kw):
+            pass
+
+        def execute(self, *args, **kwargs):
+            return numpy.ones((2, 2), dtype=int)
+
+    def test_arraylike_raises_if_shapes_are_mismatched(self, graph):
+        op = self.OpA(graph=graph)
+
+        with pytest.raises(Exception):
```
k-dominik: ugh, this is all done with asserts, at least in `ArrayLike`.  Should probably check for `AssertionError`, slightly more specific. But I guess it's all wrong that it is done with asserts in the first place.

lazyflow/slot.py
m-novikov: Misleading comment, it remained after I removed some code, fixed

lazyflow/slot.py
```diff
@@ -931,26 +931,13 @@ def __call__(self, destination=None):
...
-                            " result of correct shape. result shape is"
-                            " {} vs {}.  roi was {}".format(
-                                self.operator, result_op.shape, destination.shape, str(self.roi)
-                            )
-                        )
-                    destination = result_op
-
-                    # check that the returned value is compatible with the requested roi
-                    self.slot.stype.check_result_valid(self.roi, destination)
+                    if destination_given and id(result_op) != id(destination):
```
emilmelnikov:
```suggestion
                    if destination_given and result_op is not destination:
```

lazyflow/stype.py
```diff
@@ -232,26 +240,22 @@ def copy_data(self, dst, src):
...
+                raise InvalidResult(f"Result has shape {len(result)}, but expected {expected_shape}")
+
         elif isinstance(result, h5py.Group):
             # FIXME: this is a hack. the slot
             # OpCompressedCache.OutputHdf5 is not really array-like,
             # because it expects destinations of type h5py.Group.
             pass
         else:
-            assert False, "check_result_valid: result type is not supported"
+            raise InvalidResult(f"Result of type {type(result)} is not supported")
```
emilmelnikov:
```suggestion
            raise InvalidResult(f"Result of type {type(result).__name__} is not supported")
```

PR #310, opened 2019-06-26
Mutlilevel slots should not implement operator interface

m-novikov:
Reasons:
* Reduces recursion depth
* Simplifies reasoning about the graph structure


k-dominik:
let's say goodbye to fakeoperatorslots!

PR #313, opened 2019-07-02
Streamline op data reader

Tomaz-Vieira:
## Changed

Streamlines opDataReader ode and adds convenience path methods

## To-do

Pack Slice5D, Shape5D and friends into their own standalone module instead of adding it directly into lazyflow;


lazyflow/utility/pathHelpers.py
```diff
@@ -289,6 +291,15 @@ def isUrl(path):
     # For now, the simplest rule will work.
     return "://" in path
 
+def isRelative(path:str) -> bool:
+    return not isUrl(path) and not os.path.isabs(path)
+
+def splitPath(path:str) -> List[str]:
```
k-dominik: should maybe be called `splitPaths`

lazyflow/operators/ioOperators/opInputDataReader.py
```diff
@@ -152,21 +165,18 @@ def setupOutputs(self):
...
-        # Does this look like a relative path?
-        useRelativePath = not isUrl(filePath) and not os.path.isabs(filePath)
-
-        if useRelativePath:
-            # If using a relative path, we need both inputs before proceeding
-            if not self.WorkingDirectory.ready():
-                return
+        path_components = splitPath(self.FilePath.value)
+
+        cwd = self.WorkingDirectory.value if self.WorkingDirectory.ready() else None
```
emilmelnikov:
```python
if not self.WorkingDirectory.ready():
    return  # FIXME: this mirrors old logic but I'm not sure if it's safe

cwd = self.WorkingDirectory.value
path_components = splitPath(self.FilePath.value)
filePath = os.path.pathsep.join(make_absolute(p, cwd) for p in path_components)
```

Tomaz-Vieira: I really dislike this `WorkingDirectory` slot here -.- , but this is not really the same logic as the one in the PR; If this method only got absolute paths in its `FilePath` input, then it'd be fine even without a `ready()` `WorkingDirectory` slot.

PR #322, opened 2019-08-05
Fix priorities propagation on direct execute

m-novikov:
Volumina priorities get ignored on direct execute


k-dominik:
closing this in favor of #323

PR #323, opened 2019-08-06
Rework scheduling logic

m-novikov:


tests/testRequest.py
```diff
@@ -653,14 +653,13 @@ def onfinish(r, result):
 
         req = Request(f)
         req.notify_finished(partial(onfinish, req))
-
         req.submit()
         req.wait()
         del req
 
         # The ThreadPool._Worker loop has a local reference (next_task),
         # so wait just a tic for the ThreadPool worker to cycle back to the top of its loop (and discard the reference)
```
k-dominik: comment should be updated accordingly

lazyflow/request/threadPool.py
```diff
@@ -87,98 +101,44 @@ class _Worker(threading.Thread):
...
         # Try to get some work.
         self.state = "waiting"
-        next_task = self._get_next_job()
 
-        while not self.stopped:
-            # Start (or resume) the work by switching to its greenlet
+        logger.info("Started worker %s", self)
+
+        while True:
+            next_task = self.get_next_task()
```
k-dominik: Should we keep track of state here, too?

```suggestion
            self.state = "waiting"
            next_task = self.get_next_task()
```

lazyflow/request/threadPool.py
```diff
@@ -39,13 +40,32 @@ class ThreadPool:
...
+    def _distribute_work(self):
+        while True:
+            try:
+                task = self.unassigned_tasks.get()
+                worker = getattr(task, "assigned_worker", None)
+
+                if not worker:
+                    worker = self.ready_workers.get()
+
+                worker.job_queue.put(task)
```
emilmelnikov: Maybe use `put_nowait` just to be sure that we won't ever block?

lazyflow/request/threadPool.py
```diff
@@ -39,13 +40,32 @@ class ThreadPool:
     def __init__(self, num_workers: int):
         """Start all workers."""
         self.unassigned_tasks = queue.PriorityQueue()
+        self.ready_workers = queue.Queue()
```
emilmelnikov: :+1:

lazyflow/request/threadPool.py
```diff
@@ -39,13 +40,32 @@ class ThreadPool:
...
+                task = self.unassigned_tasks.get()
+                worker = getattr(task, "assigned_worker", None)
+
+                if not worker:
+                    worker = self.ready_workers.get()
+
+                worker.job_queue.put(task)
+                task = None
+            except Exception as e:
+                logger.exception("Exception in scheduler")
```
emilmelnikov:
```suggestion
                logger.exception(f"Unhandled exception in the scheduler thread: {e!r}")
```

lazyflow/request/threadPool.py
```diff
@@ -39,13 +40,32 @@ class ThreadPool:
...
         atexit.register(self.stop)
 
+        self._scheduler = threading.Thread(target=self._distribute_work)
+        self._scheduler.daemon = True
+        self._scheduler.start()
+
+    def _distribute_work(self):
+        while True:
+            try:
+                task = self.unassigned_tasks.get()
```
emilmelnikov: Shouldn't we stop the scheduler as well?

```python
if task is STOP:
    return

....

if worker is STOP:
    return

....

def stop(...):
    ...
    self.unassigned_tasks.put_nowait(STOP)
    self.ready_workers.put_nowait(STOP)
```

tests/testRequest.py
```diff
@@ -653,14 +653,13 @@ def onfinish(r, result):
...
-        time.sleep(0.1)
+        gc.collect()
```
emilmelnikov: Do we really need this in a test?

lazyflow/request/threadPool.py
m-novikov: It's a daemon thread and it doesn't do any actual work, so it will gracefully shut down by itself with thread pool termination. Otherwise, we would need to add timeouts to both queues gets and introduce polling to properly terminate scheduler.

lazyflow/request/threadPool.py
m-novikov: Moved `self.state = "waiting"`

lazyflow/request/threadPool.py
m-novikov: Kept the message except for the reference to exception object `{e!r}` as `logger.exception` already provides all exception information.

tests/testRequest.py
m-novikov: It broke when I didn't free the reference :) More robust test would probably try to overload thread pool with work and than check if there was any memory leakage.

lazyflow/request/threadPool.py
emilmelnikov: That's great, didn't knew that!

lazyflow/request/threadPool.py
emilmelnikov: Wouldn't putting `STOP` in both queues solve the problem?

Keeping a running thread after `ThreadPool.stop()` doesn't seem like a good behavior (it consumes resources and prevents GC for some objects), but this is probably not a big deal since we have 1 pool for the entire lifetime of a process anyway.

m-novikov: And then we need to make STOP object comparable with objects in unassigned_tasks.

emilmelnikov: It should be doable:

```python
@functools.total_ordering
class Stop:
    __lt__ = lambda _self, _other: True
    __eq__ = lambda self, other: isinstance(self, other.__class__)


STOP = Stop()
```

And we also need to fix `Request`, which should be done anyway because right now it's `__lt__` method is broken (works only with other request object):

```python
@functools.total_ordering
class Request:
    ...

    def __lt__(self, other):
        if not hasattr(other, "_priority"):
            return NotImplemented
        return self._priority < other._priority

    def __eq__(self, other):
        if not hasattr(other, "_priority"):
            return NotImplemented
        return self._priority == other._priority            
```

This is probably even better because `STOP` will appear immediately as the next element in all queues, which is what we want.

N.B.: I believe the comment about `None` in the current `Request.__lt__` is wrong: queues are populated only through `Request._wake_up` that always passes self as an argument.

m-novikov: See updated revision

emilmelnikov: Great! Explicitly passing the priority is an even better approach.

lazyflow/request/threadPool.py
```diff
@@ -37,15 +38,15 @@ class StopException(Exception):
 class QueueObject:
     __slots__ = ("obj", "exc", "priority")
 
-    def __init__(self, *, obj=None, exc=None, priority=None):
+    def __init__(self, *, obj: Any = None, exc: Exception = None, priority: Optional[List[int]] = None) -> None:
```
emilmelnikov: Maybe use a dataclass for this struct? Seems appropriate.

m-novikov: Don't really want to import dataclasses here, little benefit as I need my constructor for validation.

PR #324, opened 2019-08-20
Move increment/decrement execution count to operator

m-novikov:


m-novikov:
Also fixes issue that filter operator will run during shutdown on unready slots

Tomaz-Vieira:
I think I did just that in some lost WIP long ago... I'm glad to see it again. LGTM =)

PR #325, opened 2019-08-21
Fix paths with z5py1.5

k-dominik:
with a recent update of z5py, it now correctly returns internal paths only from `visiteditems`:

https://github.com/constantinpape/z5/commit/8ec2ff7555a034b49a2666f79e86707713c93164

requires `z5py>=1.5.1`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,4 +3,3 @@ repos:
     rev: stable
     hooks:
     - id: black
-language_version: python3.6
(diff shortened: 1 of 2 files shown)


m-novikov:
Do we care about z5py 1.4?

constantinpape:
> Do we care about z5py 1.4?

I don't think there is any reason to support anything < 1.51.
1.51 is on conda-forge, it fixes a lot of bugs and does not link to boost any longer, which makes it behave nicer as a dependency.

PR #329, opened 2019-10-07
Add factory function to create requests returning value

m-novikov:
When we have a certain function that usually returns a Request, but in some cases, when it can't do any actual work and should fallback to just returning a default value we should be able to create dummy request returning value.

Example:

```python
def foo(val):
     if val is None:
        return Request.with_value(32)

     return Request(_heavy_computational_func)
```
I moved ValueRequest from slot.py because it's functionality relevant only to request users.

### DIFF
--- a/lazyflow/request/request.py
+++ b/lazyflow/request/request.py
@@ -1575,7 +1575,7 @@ class _ValueRequest:
             destination.data[...] = ma.getdata(self.result)
             destination.mask[...] = ma.getmaskarray(self.result)
             if isinstance(self.result, ma.masked_array):
-                destination.fill_value = self.result.fill_value
+                destination.set_fill_value(self.result.fill_value)
 
         elif isinstance(destination, collections.MutableSequence) or isinstance(
             self.result, collections.MutableSequence
--- a/lazyflow/request/request.py
+++ b/lazyflow/request/request.py
@@ -22,6 +22,7 @@ from future.utils import raise_with_traceback
 # 		   http://ilastik.org/license/
 ###############################################################################
 # Built-in
+import collections
 import sys
 import heapq
 import functools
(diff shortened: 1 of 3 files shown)


Tomaz-Vieira:
As far as the code goes, LGTM
Still, we have a `ValueRequest` class that skips all of the scheduling machinery in Lazyflow. Are you intentionally avoiding that?

m-novikov:
Nope, will use it.

m-novikov:
Regarding removal of masked array issue reference. Copying masked array works now, but it doesn't copy fill_value, so we still need this if branch.

lazyflow/request/request.py
```diff
@@ -1522,3 +1535,54 @@ def clean(self):
...
+
+    def clean(self):
+        self.result = None
+
+    def writeInto(self, destination):
+        if isinstance(destination, ma.masked_array):
+            destination.data[...] = ma.getdata(self.result)
+            destination.mask[...] = ma.getmaskarray(self.result)
+            if isinstance(self.result, ma.masked_array):
+                destination.fill_value = self.result.fill_value
```
k-dominik: have you tested this code? From the docs i take that `.fill_value` returns the fill value and `.set_fill_value()` is used to set it:
 * https://docs.scipy.org/doc/numpy-1.14.1/reference/maskedarray.baseclass.html#numpy.ma.MaskedArray.fill_value

we currently use numpy `1.14`
`fill_value` as a property is only available as of numpy `1.17.2` (`set_fill_value()` is kept there for compatibility)

m-novikov: > we currently use numpy 1.14
fill_value as a property is only available as of numpy 1.17.2 (set_fill_value() is kept there for compatibility)

I assume it is there just wasn't documented as it worked on even older versions of numpy. Added test.

k-dominik: yeah sorry, I read the code wrong. But really great that it's tested now :)

PR #334, opened 2020-02-11
Reduce number of warnings

k-dominik:
On a quest in reducing noise in tests, and in ilastik in general I've tackled most warnings issued by dependencies. Warning count went down from 188 to 24 when running the tests.

Warnings removed:

* `numpy`:
  * DeprecationWarning: The binary mode of fromstring is deprecated
  * FutureWarning: Conversion of the second argument of issubdtype from `dtype` to `np.generic` is deprecated. In future, it will be treated as `np.object_ == np.dtype(dtype).type`.
  * DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
* `h5py`:
  * H5pyDeprecationWarning: dataset.value has been deprecated. Use dataset[()] instead.

I'm doing the same in the other repos while I'm at it. Now it's easier to concentrate on our own warnings...

btw. `numpy.array_equal` (and it's sister `numpy.array_equiv`) are awesome, didn't know them before. Those functions eat almost anything

### DIFF
--- a/lazyflow/operators/opSlicedBlockedArrayCache.py
+++ b/lazyflow/operators/opSlicedBlockedArrayCache.py
@@ -138,8 +138,8 @@ class OpSlicedBlockedArrayCache(Operator, ObservableCache):
         ram_per_pixel = 0
         if self.Output.meta.dtype == object or self.Output.meta.dtype == numpy.object_:
             ram_per_pixel = sys.getsizeof(None)
-        elif numpy.issubdtype(self.Output.meta.dtype, numpy.dtype):
-            ram_per_pixel = self.Output.meta.dtype().nbytes
+        elif numpy.issubdtype(self.Output.meta.dtype, numpy.generic):
+            ram_per_pixel = numpy.dtype(self.Input.meta.dtype).itemsize
 
         tagged_shape = self.Output.meta.getTaggedShape()
         if "c" in tagged_shape:
--- a/lazyflow/slot.py
+++ b/lazyflow/slot.py
@@ -1108,7 +1108,6 @@ class Slot(object):
                             # in which case we assume the values are different
                             same = False
 
-                        assert isinstance(same, bool)
                     changed = not same
 
             if changed:
--- a/lazyflow/slot.py
+++ b/lazyflow/slot.py
@@ -1102,13 +1102,13 @@ class Slot(object):
                     same = value is self._value
                     if not same:
                         try:
-                            same = value == self._value
+                            same = numpy.array_equal(value, self._value)
                         except ValueError:
                             # Some values can't be compared with __eq__,
                             # in which case we assume the values are different
                             same = False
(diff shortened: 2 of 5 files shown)


lazyflow/slot.py
```diff
@@ -1102,13 +1102,13 @@ def setValue(self, value, notify=True, check_changed=True, extra_meta={}):
...
-                            same = value == self._value
+                            same = numpy.array_equal(value, self._value)
                         except ValueError:
                             # Some values can't be compared with __eq__,
                             # in which case we assume the values are different
                             same = False
-                        if isinstance(same, (numpy.ndarray, TinyVector)):
-                            same = same.all()
+
+                        assert isinstance(same, bool)
```
Tomaz-Vieira: I can't see any path where `type(same) != bool`... is this still necessary?

k-dominik: me neither, but I am paranoid :ghost:

lazyflow/operators/opSlicedBlockedArrayCache.py
```diff
@@ -138,8 +138,8 @@ def setupOutputs(self):
         ram_per_pixel = 0
         if self.Output.meta.dtype == object or self.Output.meta.dtype == numpy.object_:
             ram_per_pixel = sys.getsizeof(None)
-        elif numpy.issubdtype(self.Output.meta.dtype, numpy.dtype):
-            ram_per_pixel = self.Output.meta.dtype().nbytes
+        elif numpy.issubdtype(self.Output.meta.dtype, numpy.generic):
+            ram_per_pixel = numpy.dtype(self.Input.meta.dtype).itemsize
```
Tomaz-Vieira: I see this `ram_per_pixel` value is now based on the size of `self.Input.dtype` rather than `self.Output.dtype`. Is it correct to have `ram_per_pixel` be base on `Input` some times and on `Output` other times?

k-dominik: that was a type, good catch!

Tomaz-Vieira:
Some heroic work there. I like less pollution in my logs =D

lazyflow/operators/opSlicedBlockedArrayCache.py
k-dominik: on the other hand, the cache doesn't change data-type. So technically it doesn't make a difference. The whole surrounding code also only checks the input. So for now I'd rather leave it like this than going for either a) inconsistency with the surrounding code, or b) polluting this PR with unrelated changes. Are you ok with that?

Tomaz-Vieira: Sure. Let's keep it simple and reviewable. We'll get back to it later =)

k-dominik:
I'm closing this and will do a cleanup for both ilastik + lazyflow once the repos are merged. It's nicer that way imo

PR #343, opened 2020-03-17
Test refactor

m-novikov:
Test naming convention is from CPython repo, all packages prefixed with test_ as modules.
In the future test folder would also contain test_ilastik as it would be merged together.


m-novikov:
There was an issue with brittle test resulted in zero division error because of memory consumption. I set the minimal level to 4096 pixels in a block (1, 64, 64).
The `testOpH5N5WriterBigDataset.py ` is unchanged.

tests/test_lazyflow/test_operators/test_ioOperators/testOpH5N5WriterBigDataset.py
```diff
@@ -0,0 +1,197 @@
...
+        assert h5_success
+        assert n5_success
+
+        hdf5File.close()
+        n5File.close()
+
+        # Check the file.
+        hdf5File = h5py.File(h5_path, "r")
+        n5File = z5py.File(n5_path, "r")
+        h5_dataset = hdf5File[datasetInternalPath]
```
Tomaz-Vieira: Does writing "lots of tiny blocks" also mean that the block size within the H5 (or N5) should reflect that size?
If we have any expectations from the stored block ize, than maybe we could if `dataset.chunks` match the shape we're expecting

m-novikov: I suppose this is a valid idea worth investigating, but this file is mostly the same as [original](https://github.com/ilastik/lazyflow/blob/master/tests/testOpH5N5WriterBigDataset.py) so I don't plan to introduce any changes to it.

lazyflow/roi.py
```diff
@@ -751,7 +751,7 @@ def determine_optimal_request_blockshape(
     ideal_blockshape = numpy.asarray(ideal_blockshape)
 
     target_block_volume_bytes = available_ram // num_threads
-    target_block_volume_pixels = target_block_volume_bytes // ram_usage_per_requested_pixel
+    target_block_volume_pixels = max(4096, target_block_volume_bytes // ram_usage_per_requested_pixel)
```
Tomaz-Vieira: Is this the code that could output 0 and cause that "division by 0" error? Seems like this would happen only if `available_ram` was set to 0. But in that case, shouldn't this function throw an exception? Or just assume that it was a caller error?

m-novikov: No, it's integer division so for example ```100 // 300 == 0```

tests/test_lazyflow/test_operators/test_ioOperators/testOpH5N5WriterBigDataset.py
m-novikov: Reverted changes to this test, they were made during investigating reason behind zero division.

lazyflow/roi.py
Tomaz-Vieira: Ah, of course!

PR #344, opened 2020-03-18
Cache mem manager

m-novikov:
Make CacheMemoryManager testable. Split implementation and singleton instantiation. 
Otherwise, tests break if run together with ilastik tests.

### DIFF
--- a/lazyflow/operators/cacheMemoryManager.py
+++ b/lazyflow/operators/cacheMemoryManager.py
@@ -28,24 +28,23 @@ import threading
 import weakref
 import functools
 import atexit
+import warnings
 
 
 # lazyflow
 from lazyflow.utility import OrderedSignal
-from lazyflow.utility import Singleton
 from lazyflow.utility import log_exception
 from lazyflow.utility import Memory
 
 
 import logging
-from future.utils import with_metaclass
 
 logger = logging.getLogger(__name__)
 
 default_refresh_interval = 10
 
 
-class CacheMemoryManager(with_metaclass(Singleton, threading.Thread)):
+class _CacheMemoryManager(threading.Thread):
     """
     class for the management of cache memory
 
@@ -55,16 +54,11 @@ class CacheMemoryManager(with_metaclass(Singleton, threading.Thread)):
     class. See the definition of the cache interfaces (opCache.py) to
     get an overview over the possible caches.
 
-    Usage:
-    This manager is a singleton - just call its constructor somewhere
-    and you will get a reference to the *only* running memory management
-    thread.
-
     Interface:
     The manager provides a signal you can subscribe to::
 
         def writeFunction(x): print("total mem: {}".format(x))
-            mgr = CacheMemoryManager()
+            mgr = _CacheMemoryManager()
             mgr.totalCacheMemory.subscribe(writeFunction)
 
     which emits the size of all observable caches, combined, in regular
@@ -73,7 +67,8 @@ class CacheMemoryManager(with_metaclass(Singleton, threading.Thread)):
     The update interval (for the signal and for automated cache release)
     can be set with a call to a class method::
 
-        CacheMemoryManager().setRefreshInterval(5)
+        cache_mem_manager = _CacheMemoryManager()
+        cache_mem_manager.setRefreshInterval(5)
 
     the interval is measured in seconds. Each change of refresh interval
     triggers cleanup.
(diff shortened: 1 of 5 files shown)


lazyflow/operators/opCache.py
```diff
@@ -54,8 +54,11 @@ class MyCachingOperator(Cache, Operator):
     cleanup could occur while the cache is still under construction!
     """
 
-    def registerWithMemoryManager(self):
-        manager = CacheMemoryManager()
+    def registerWithMemoryManager(self, memory_manager=None):
```
k-dominik: what is the use-case for passing a memory manager? Just testing, or do you also have something else in mind?

lazyflow/operators/cacheMemoryManager.py
```diff
@@ -280,3 +275,28 @@ def enable(self):
             self._disabled = False
         with self._condition:
             self._condition.notifyAll()
+
+
+_cache_memory_manager = _CacheMemoryManager()
+totalCacheMemory = _cache_memory_manager.totalCacheMemory
+
+
+def CacheMemoryManager():
```
Tomaz-Vieira: Tests seem to be passing, but are you sure no one is doing stuffthat relies on CacheMemoryManager being a class like:
`CacheMemoryManager.__class__`
or
```python3
class MyOtherCacheManager(CacheMemoryManager):
    ...
```
?

I've found these log configurations in `ilastik/ilastik/ilastik_logging/default_config.py:156 `:

`"lazyflow.operators.cacheMemoryManager.CacheMemoryManager": {"level": "INFO"},`

and similar in `ilastik/ilastik/ilastik_logging/logging_config.json`. Do you know if they will still work if CachememoryManager is no longer a class?

lazyflow/operators/opCache.py
m-novikov: Just testing, do you we should change this somehow maybe add underscore to indicate that it is an argumetn for internal use only?

lazyflow/operators/cacheMemoryManager.py
k-dominik: so just for the logging, we can safely remove the trailing `.CacheMemoryManager`. ilastik/ilastik of course needs some changes, too. But I haven't seen any subclassing or things like that.

m-novikov: Even now there is no logger with this name :) Only existing logger is "lazyflow.operators.cacheMemoryManager"

tests/test_lazyflow/conftest.py
```diff
@@ -16,6 +16,11 @@ def pytest_exception_interact(node, call, report):
             report.sections.append(("Operator Stack", formatted_stack))
 
 
+@pytest.hookimpl()
+def pytest_runtest_setup(item):
```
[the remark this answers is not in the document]
m-novikov: Added comment see https://github.com/ilastik/lazyflow/pull/344/commits/b7daf4f5198c7c8d8d3df4b3ebdd8c2027e0e0c0

lazyflow/operators/opCache.py
k-dominik: could we maybe make this a private method (with an underscore) and a mandatory `memory_manager` argument and a public one that just passes `cacheMemoryManager` to the private one?

m-novikov: Problem with this approach that it's called from constructors of Op*Cache. See usage in tests https://github.com/ilastik/lazyflow/blob/b7daf4f5198c7c8d8d3df4b3ebdd8c2027e0e0c0/tests/test_lazyflow/test_operators/testCacheMemoryManager.py#L174 so as a result each cache would need to know to call this private method or we would have two managers with different policies managing the cache.

lazyflow/operators/cacheMemoryManager.py
m-novikov: > Tests seem to be passing, but are you sure no one is doing stuffthat relies on CacheMemoryManager being a class like:
> `CacheMemoryManager.__class__`
> or
> 
> ```python
> class MyOtherCacheManager(CacheMemoryManager):
>     ...
> ```
Of this I am sure, run tests in ilastik and inspected the code in volumina it's not used at all

lazyflow/operators/opCache.py
k-dominik: how about not exposing this at all and using pytests monkeypatch functionality to inject a cachememorymanager during testing?

m-novikov: > how about not exposing this at all and using pytests monkeypatch functionality to inject a cachememorymanager during testing?

Yep, great idea :)

emilmelnikov: Is implicit monkey-patching really better than adding an explicit argument in this case? `OpCache` might as well work with anything that looks like a memory manager, not just a single global one.

k-dominik: unless that there is a good reason for using a _specific_ different CacheMemoryManager I'd prefer to send a clear message to user of this class.

emilmelnikov: Testing seems like a good reason since it frequently uncovers hidden dependencies. It seems strange to monkey-patch something that is under our control.

As for the clear message, we can add a docstring; if someone wants to use this param, they are expected to read docs anyway. Also, if you decide to leave this as a param, it might be a good idea to make it keyword-only.

m-novikov: I agree with monkey patching approach because figuring out the best way to do dependency injection is out of the scope of this pull-request. The purpose of this pull-requests is only to enable to run lazyflow tests together with ilastik.

